PR #44, opened 2020-11-02
Pre commit linting

l0b0:
Would this be a useful addition to the pre-commit linting we already have in place?

Also wondering if we could merge the dev env linting and CI linting by running `pre-commit run --all-files` in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,209 +1,221 @@
+---
 name: Build
-on: [push]
+on: [push]  # yamllint disable-line rule:truthy
 
 jobs:
-  linter:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python: ["3.6", "3.7"]
+    linter:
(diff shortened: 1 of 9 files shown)


imincik (commented):
@l0b0 , looks good to me. Until I had tuned all linting in my editor I had a lot of failed CI runs because of stupid linting error. This can save a lot of time.

Are you able to add pylint commit hook as well to have them all ?

Thank you

l0b0:
> Are you able to add pylint commit hook as well to have them all ?

Done.

This led to a better understanding of pre-commit since pylint has to run in the local Virtualenv to validate the imports in the code (Black and isort just format the code without actually ever following imports). But running Git hooks within the local Virtualenv has other benefits, like avoiding version drift between the hooks and the requirements file and enabling checksumming the package installs, so I just converted all of the hooks to use the local Virtualenv.

imincik (approved):
Thanks @l0b0

PR #45, opened 2020-11-04
test(lint): Add Dependabot support for GitHub Actions

l0b0:
Based on https://github.com/linz/template-python-hello-world


imincik (approved):
Thanks @l0b0 . Looks good.

PR #47, opened 2020-11-04
refactor(python): Use Poetry

l0b0:


pyproject.toml
```diff
@@ -8,3 +8,38 @@ use_parentheses = true
 include_trailing_comma = true
 multi_line_output = 3
 force_grid_wrap = 0
+
+[tool.poetry]
+name = "geospatial-data-lake"
+version = "0.1.0"
+description = "Central storage, management and access for important geospatial datasets developed by LINZ"
+authors = [
+    "Bill Nelson <billgeo@users.noreply.github.com>",
```
imincik, line 17: Incorrect Bill's email - should be `Bill M. Nelson <bmnelson@linz.govt.nz>`

imincik (dismissed):
Thank you @l0b0 , I am very happy to switch to poetry. I left one comment about Bill's email. Otherwise looks good.

pyproject.toml
l0b0: Thank you; changed. I got all the emails out of `git log --format='%aN <%aE>' | sort --unique`, so I wasn't sure whether I should "de-anonymize" Bill that way :)

imincik:
@l0b0 , would you please rebase this PR with current master ? I am very interested how it is going to look like. Thanks.

imincik (commented):
I am not sure if moving `backend/endpoints/datasets/requirements.txt` to `backend/requirements.txt` is good idea, because we need separate `requirements.txt` file with different content for each Lambda function. New endpoint lambda functions will be created in `backend/endpoints` dir and all of them will require their own `requirements.txt` file. Also there will be dozens of other Lambda functions which will belong to data validation and import process (all of them again with different requirements).

In general PR looks really good. Just need to solve those requirements files. Thanks

imincik (approved):
Thanks @l0b0 . Looks good now.

PR #53, opened 2020-11-12
style: Add shellcheck

l0b0:


imincik (changes requested):
Hi @l0b0 , i see some `shellcheck` unrelated changes in this PR in `.github/dependabot.yml`. If it's not needed I would separate it to another PR. Thanks

imincik (approved):
Good. Thanks.

PR #56, opened 2020-11-14
feat(infra): state machine workflow for datasets validation

imincik:
Implements Lambda State Machine workflow for datasets validation during Dataset Version creation according [Datasets Version Endpoint Detailed Design](https://confluence.linz.govt.nz/display/LGDL/Dataset+Version+Endpoint)

This PR implements includes dummy implementation of all workflow steps. Also includes new AWS stacks: 

* networking - VPC required for AWS Batch (ECS) cluster
* processing - State Machine resources


imincik:
@l0b0 , one problem with this PR I see right now is that we have duplicated Lambda functions bundling script (`backend/dataset_import/bundle.bash`). I am not sure if we want to solve it in this PR or in the new one. What do you think ?

l0b0:
> @l0b0 , one problem with this PR I see right now is that we have duplicated Lambda functions bundling script (`backend/dataset_import/bundle.bash`). I am not sure if we want to solve it in this PR or in the new one. What do you think ?

Shouldn't `dataset_import` be in the `endpoints` directory? The idea with the bundle.bash is that it can be called with the endpoint name, so we don't need to duplicate the script to use it for all the endpoints.

imincik:
@l0b0 , I wanted to have separate directory for `endpoint` lambda functions. They are main API endpoints and it looks to me that they might deserve separate directory. As we go, I guess there would be some need to have even more dirs for different kinds of Lambda functions - just to avoid one big directory with mess of functions of different kind.

If you agree, then we need to move bundling shell script one directory higher and use it then for all Lambdas. The question is if we should do it in this PR or do in separate PR. Personally, I would be happier to do it in another PR.

backend/processing/bundle.bash
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
l0b0, line 1: As discussed, we should probably reuse the existing bundle.bash script. It should be pretty easy if you move it up a level and use `basename`/`dirname` on `$1`.

backend/processing/check_files_checksums/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ubuntu:18.04
```
l0b0, line 1: We should use Ubuntu 20.04 - it's the latest LTS, so it'll be supported for longer.

backend/processing/check_files_checksums/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+RUN apt update && apt install -y python3
```
l0b0, line 3: We should [clean up the APT cache after installing](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) to minimize the size of the Docker layer.

backend/processing/check_files_checksums/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+COPY task.py /opt/task.py
+RUN chmod +x /opt/task.py
```
l0b0, line 5: If you make task.py executable *locally* you shouldn't need to run this.

backend/processing/check_files_checksums/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+ENTRYPOINT ["python3", "/opt/task.py"]
```
l0b0, line 6: If you add a shebang line like `#!/usr/bin/env python3` you can simplify this to just `ENTRYPOINT ["/opt/task.py"]`.

backend/processing/check_files_checksums/task.py
```diff
@@ -0,0 +1,10 @@
+"""
+TODO: add docstring here.
```
l0b0, line 2: We don't need these anymore.

backend/processing/check_flat_directory_structure/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ubuntu:18.04
```
l0b0, line 1: This file is identical to the other Dockerfile. Can we use the same Dockerfile for all of the processing steps?

backend/processing/content_iterator/task.py
```diff
@@ -0,0 +1,33 @@
+"""
+TODO: add docstring here.
+
+"""
+
+
+def lambda_handler(event, _context):  # pylint:disable=unused-argument
```
l0b0, line 7: You don't need this pylint annotation here.

backend/processing/validation_failure/task.py
```diff
@@ -0,0 +1,9 @@
+"""
+TODO: add docstring here.
+"""
+
+
+def lambda_handler(event, _context):  # pylint:disable=unused-argument
```
l0b0, line 6: If you add an underscore to the start of the first argument you shouldn't need the `pylint` annotation.

infra/app.py
```diff
@@ -8,17 +8,37 @@
 
 from aws_cdk import core
 from datalake.datalake_stack import DataLakeStack  # pylint: disable=import-error
+from datalake.networking_stack import NetworkingStack  # pylint: disable=import-error
```
l0b0, line 11: How do we avoid having to annotate all these lines? Do we need to move some files?

infra/app.py
```diff
@@ -8,17 +8,37 @@
...
-    env={"region": "ap-southeast-2"},
+    "networking",
+    stack_name=f"geospatial-data-lake-networking-{ENV}",
+    env={"region": os.environ["CDK_DEFAULT_REGION"], "account": os.environ["CDK_DEFAULT_ACCOUNT"]},
 )
 
+processing = ProcessingStack(
+    app,
+    "processing",
+    stack_name=f"geospatial-data-lake-processing-{ENV}",
```
l0b0, line 29: We can add `ENV` to the stack name within the stack __init__, to avoid effectively passing this value twice.

infra/datalake/datalake_stack.py
```diff
@@ -9,10 +9,10 @@ class DataLakeStack(core.Stack):
     """Data Lake stack definition."""
 
     # pylint: disable=redefined-builtin,too-many-locals
-    def __init__(self, scope: core.Construct, id: str, **kwargs) -> None:
+    def __init__(self, scope: core.Construct, id: str, deploy_env, **kwargs) -> None:
         super().__init__(scope, id, **kwargs)
 
-        ENV = self.stack_name.split("-")[-1]
+        ENV = deploy_env
```
l0b0, line 15: We should not override "constants" within methods. That basically makes them globals. We might as well refer to `deploy_env` within this method.

infra/datalake/networking_stack.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+from aws_cdk import aws_ec2, core
+from aws_cdk.core import Tags
+
+
+class NetworkingStack(core.Stack):
+    """Data Lake networking stack definition."""
+
+    # pylint: disable=redefined-builtin,too-many-locals
+    def __init__(self, scope: core.Construct, id: str, **kwargs) -> None:
```
l0b0, line 12: We should not redefine builtins. A common workaround is basically Hungarian notation: add the context to the name, such as `stack_id`.

infra/datalake/processing_stack.py
```diff
@@ -0,0 +1,309 @@
...
+                            # pylint:disable=no-member
+                            image=aws_lambda.Runtime.PYTHON_3_6.bundling_docker_image,
+                            command=["backend/processing/bundle.bash", f"{task}"],
+                        ),
+                    ),
+                )
+
+                step_tasks[task] = aws_stepfunctions_tasks.LambdaInvoke(
+                    self,
+                    f"{task}",
```
l0b0, line 184: This could be just `task,`

infra/datalake/processing_stack.py
```diff
@@ -0,0 +1,309 @@
...
+                    "type.$": "$.type",
+                    "metadata_url.$": "$.metadata_url",
+                }
+                job_payload_data_parallel = {"first_item.$": "$.content.first_item"}
+                job_payload_single = aws_stepfunctions.TaskInput.from_object(job_payload_data)
+                job_payload_parallel = aws_stepfunctions.TaskInput.from_object(
+                    {**job_payload_data, **job_payload_data_parallel}
+                )
+
+                if creation_tasks[task]["parallel"]:
```
l0b0, line 235: `creation_tasks[task]` is used so many times, we should probably rename `task` to `task_name` and pull out `task = creation_tasks[task_name]`.

infra/datalake/processing_stack.py
```diff
@@ -0,0 +1,309 @@
...
+                            step_tasks["success"],
+                        )
+                        .otherwise(step_tasks["validation_failure"])
+                    ),
+                )
+            )
+        )
+
+        # state machine
+        creation_process = aws_stepfunctions.StateMachine(  # pylint:disable=unused-variable
```
l0b0, line 305: Why not just don't create the variable, to avoid `pylint` annotations? In the vast majority of cases, whenever Pylint complains I would change the code rather than adding an annotation.

pyproject.toml
```diff
@@ -29,6 +29,12 @@ python = "^3.6"
 "aws-cdk.aws-iam" = {version="*", optional=true}
 "aws-cdk.aws-dynamodb" = {version="*", optional=true}
 "aws-cdk.aws-lambda" = {version="*", optional=true}
+"aws-cdk.aws-stepfunctions" = {version="*", optional=true}
```
l0b0, line 32: Can you please sort the package names alphabetically within this and the next block? It just makes it easier to scan the whole thing for a specific name.

backend/processing/bundle.bash
```diff
@@ -13,16 +13,17 @@ fi
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 work_dir="$(mktemp --directory)"
+asset_root='/asset-output'
```
l0b0, line 16: Good idea, but since this doesn't actually fix a bug I'd change the commit message to something like "refactor: Pull out variable"

backend/processing/bundle.bash
```diff
@@ -13,16 +13,17 @@ fi
...
 
 if [ -f "${script_dir}/${1}/requirements.txt" ]; then
     python -m venv "${work_dir}/.venv"
     # shellcheck source=/dev/null
     . "${work_dir}/.venv/bin/activate"
     python -m pip install --upgrade pip
 
-    pip install --requirement="${script_dir}/${1}/requirements.txt" --target=/asset-output
+    pip install --requirement="${script_dir}/${1}/requirements.txt" --target=${asset_root}
```
l0b0, line 24: `-target=${asset_root}` should be quoted now that there's a variable in that string.

.github/workflows/ci.yml
```diff
@@ -134,9 +134,9 @@ jobs:
               run: |
                   poetry run pytest backend/tests/
 
-            - name: Destroy AWS stack used for testing
+            - name: Destroy AWS stacks used for testing
               run: |
-                  poetry run cdk destroy --force geospatial-data-lake
+                  poetry run cdk destroy --force processing datalake  # don't drop networking
```
l0b0, line 139: If you're not going to drop networking it would be good to explain *why.*

backend/processing/check_files_checksums/Dockerfile
imincik: Fixed in 3476dd6a758df5ad18cb67bf4fce299420dc0f6c

backend/processing/check_files_checksums/Dockerfile
imincik: Done in de4abcf61c542be7bacf2e11517acde1a92063b2 . Thanks

backend/processing/check_files_checksums/Dockerfile
imincik: Do you mean assigning executable permissions for script in source code ? I don't like this approach.

backend/processing/check_files_checksums/Dockerfile
imincik: Done in d1cf6350db2a444ec180d0f6eec2e2bbd95f5062 . Thanks, I was missing shebang in one script.

backend/processing/check_files_checksums/task.py
imincik: Fixed in a954b72ed0860ce9120aff8d3f1457456f2932c3

backend/processing/check_flat_directory_structure/Dockerfile
imincik: Yes it is the same, but currently I don't see any configuration in [aws_batch.JobDefinition](https://docs.aws.amazon.com/cdk/api/latest/python/aws_cdk.aws_batch/JobDefinition.html) which will allow us to use single Docker file

backend/processing/content_iterator/task.py
imincik: Done in 599f367570528d25a8e23b358e86182fe62eca96

backend/processing/validation_failure/task.py
imincik: Done in 599f367570528d25a8e23b358e86182fe62eca96

infra/app.py
imincik: No idea. This is layout created by CDK.

infra/datalake/datalake_stack.py
imincik: Fixed in f62ceb21ca87043e481e3eff95beadabd334daba

infra/datalake/networking_stack.py
imincik: This code was created by CDK at project initialization and I see most of examples on web using the same.

infra/datalake/processing_stack.py
imincik: Fixed in d9d1f2b817315e64a8669f7f3680019e23518083

infra/datalake/processing_stack.py
imincik: Because it will be needed in next stages of development.

pyproject.toml
imincik: Fixed in 7927d78cd4f574a42f780e1ea72dd2d77a8a4bb5

backend/processing/bundle.bash
imincik: Good point about commit messages in general. I think I would squash all commits before merge to master anyway, so suggest to leave it as is for now.

backend/processing/bundle.bash
imincik: Since 49287826c0b951828565effd6d92cb237f139e08 we use single Lambda bundling script

backend/processing/bundle.bash
imincik: Since 49287826c0b951828565effd6d92cb237f139e08 we use single Lambda bundling script

infra/datalake/processing_stack.py
imincik: Done in c16a3329472d71183fdbf0dbaa4928125a8ac93a

.github/workflows/ci.yml
imincik: I added comment in c16a3329472d71183fdbf0dbaa4928125a8ac93a . I think this is best I can do. We don't have any "how to add new stack" docs and I don't think it is needed.

imincik:
@l0b0 , I addressed most of your comments. There are just few which I can't solve or have some different opinion. Thanks for good review.

infra/datalake/networking_stack.py
l0b0: That's fine, it just means CDK is doing something bad (IMO) in this specific case. We should still change it.

infra/datalake/processing_stack.py
l0b0: I'll resolve this, but we should avoid building things we don't yet need (YAGNI) in general. Any work we do which is not connected to *existing* code is a waste until that code is connected to something, and increases the risk of premature design.

pyproject.toml
```diff
@@ -52,6 +58,12 @@ cdk = [
     "aws-cdk.aws-iam",
     "aws-cdk.aws-dynamodb",
     "aws-cdk.aws-lambda",
+    "aws-cdk.aws-stepfunctions",
+    "aws-cdk.aws-stepfunctions_tasks",
+    "aws-cdk.aws-ecr",
+    "aws-cdk.aws-ecr_assets",
+    "aws-cdk.aws-ecs",
+    "aws-cdk.aws-ec2",
```
l0b0, line 66: We should sort these too, as [mentioned](https://github.com/linz/geospatial-data-lake/pull/56#discussion_r534509644).

backend/processing/check_files_checksums/Dockerfile
imincik: I like more approach of setting it during deployment exactly how it should be (setting executable bit for u, g, o) with correct permissions as well.

pyproject.toml
imincik: Should be fixed in 432e847b74c4d8c7184ffb3361a442696a726759

backend/processing/check_files_checksums/Dockerfile
imincik: Can we just leave it as is please ?

infra/datalake/networking_stack.py
imincik: Should be fixed in 767304e77510a3f0271ada38e542c66dac0007ed

l0b0 (dismissed):

l0b0 (approved):

l0b0 (approved):

PR #60, opened 2020-11-16
style: Simplify getting environment variable with default

l0b0:


imincik (approved):
Yes, thank you @l0b0 .

PR #66, opened 2020-11-23
style: Don't require docstrings

l0b0:


imincik:
Why we don't want docstrings ? I saw this configuration in Topo code as well. What's the reason ?

l0b0:
> Why we don't want to have docstrings ? I saw this configuration in Topo code as well. What's the reason ?

Good question! I'll try to summarise:

1. Comments can get out of date. Code, as long as it's reached (or at least reachable) during execution, can't be out of date.
1. Comments should explain *why* the code is the way it is, not how the code works. If there is no "why" to explain, enforcing a comment is counter-productive.
1. A comment is *local* information: when I see some code I don't immediately understand completely, I have to go hunting for a relevant comment to explain it. That comment might be at the end of the line I'm reading, it could be above it, it could be at the start of the class, at the top of the file, it could have shifted a few lines if someone inserted some code between the comment and the relevant code, or it could even be below the relevant code or in a completely different place after a refactor. 
1. Since comments are not code, they have to be refactored manually.
1. After 10+ years of working in teams using minimal comments, I must confess comments are the last place I look for information. Basically the only time I end up adding a comment is if it's required (such as for generating API documentation, although there are better ways such as type hints for doing that now) or if I'm unable to make the code sufficiently clear (such as workarounds for third party bugs with links to relevant bug reports).

tl;dr: Comments are IMO rarely useful, and enforcing them on everything creates unnecessary busywork and unhelpful noise in the code.

imincik (approved):
I am ok with that.

PR #68, opened 2020-11-23
feat: Tag acceptance tests as such

l0b0:


imincik (approved):
Nice feature. I never used it.

PR #70, opened 2020-11-23
feat: Unlock shellcheck-py version

l0b0:


imincik (approved):
No problem with this. Thanks

PR #74, opened 2020-11-25
feat: Use poetry extras for subprojects

l0b0:


pyproject.toml
```diff
@@ -46,6 +46,20 @@ shellcheck-py = "^0.7.1"
...
+[tool.poetry.extras]
+cdk = [
+    "aws-cdk.aws-s3",
+    "aws-cdk.aws-iam",
+    "aws-cdk.aws-dynamodb",
+    "aws-cdk.aws-lambda",
+    "awscli",
+    "cattrs",
+]
+datasets = [
```
imincik, line 58: * can you please rename `datasets` to `datasets` endpoint ?

imincik:
When running `cdk synth` I've got following error:

```
ERROR: Invalid requirement: 'Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.' (from line 1 of /dev/fd/63)
```

So you probably need to add updated poetry lock file.

pyproject.toml
```diff
@@ -46,6 +46,20 @@ shellcheck-py = "^0.7.1"
...
+[tool.poetry.extras]
+cdk = [
+    "aws-cdk.aws-s3",
+    "aws-cdk.aws-iam",
+    "aws-cdk.aws-dynamodb",
+    "aws-cdk.aws-lambda",
+    "awscli",
+    "cattrs",
+]
+datasets_endpoint = [
```
imincik, line 58: Any reason why not to use hyphen datasets-endpont? Hyphen is already used in this file (see: tool.poetry.dev-dependencies)

l0b0: Hmm, good point. Done.

README.md
```diff
@@ -83,3 +83,11 @@ Developed by [Land Information New Zealand](https://github.com/linz)
      ```
      $ black . --check --diff
      ```
+
+To add a development-only package: `poetry add --dev PACKAGE='*'`
+
+To add a production package:
+
+1. Install the package using `poetry add --optional PACKAGE='*'`.
```
imincik, line 91: Copy/paste error in line numbers ?

l0b0: That's just how Markdown works: Any number before the dot results in the list being rendered as a numbered list, starting at 1. It's pretty common to make all the numbers 1 rather than getting a bigger diff when adding and removing entries, and dealing with re-numbering all of them.

imincik: ah, I didn't know. I had a feeling that it might be something like that.

imincik (approved):
Thanks Victor

PR #76, opened 2020-11-26
test: Tweak tests

l0b0:
I'd recommend going through this commit by commit, otherwise the full diff is a bit of a mess.


imincik (approved):
Big thumbs up @l0b0 !

PR #79, opened 2020-12-01
feat: Validate metadata

l0b0:
Based on https://github.com/linz/geospatial-data-lake/pull/132

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,7 @@ jobs:
             - uses: actions/checkout@v2
               with:
                   fetch-depth: 0  # this is to enable gitlint to check all PR commit messages
+                  submodules: 'true'
 
             - name: Use Python ${{ matrix.python }}
               uses: actions/setup-python@v2.1.4
@@ -31,7 +32,8 @@ jobs:
               run: |
                   python -m pip install --upgrade pip
                   python -m pip install poetry
-                  python -m poetry install --extras='cdk datasets-endpoint'
+                  python -m poetry install \
+                      --extras='cdk datasets-endpoint check-stac-metadata-job'
 
             - name: Check last commit message
               if: github.event_name == 'push'
@@ -63,6 +65,8 @@ jobs:
 
         steps:
             - uses: actions/checkout@v2
+              with:
+                  submodules: 'true'
 
             - name: Use Python ${{ matrix.python }}
               uses: actions/setup-python@v2.1.4
@@ -81,7 +85,8 @@ jobs:
               run: |
                   python -m pip install --upgrade pip
                   python -m pip install poetry
-                  python -m poetry install --extras='cdk datasets-endpoint'
+                  python -m poetry install \
+                      --extras='cdk datasets-endpoint check-stac-metadata-job'
 
             - name: Install AWS CLI
               run: sudo apt install awscli
@@ -157,6 +162,8 @@ jobs:
 
         steps:
             - uses: actions/checkout@v2
+              with:
+                  submodules: 'true'
 
             - name: Use Python ${{ matrix.python }}
               uses: actions/setup-python@v2.1.4
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "datalake/backend/processing/check_stac_metadata/stac-spec"]
+	path = datalake/backend/processing/check_stac_metadata/stac-spec
+	url = https://github.com/radiantearth/stac-spec.git
--- a/datalake/processing_stack.py
+++ b/datalake/processing_stack.py
@@ -15,6 +15,8 @@ from aws_cdk import (
 )
 from aws_cdk.core import Tags
 
+JOB_DEFINITION_SUFFIX = "_job"
+
 
 class ProcessingStack(core.Stack):
     """Data Lake processing stack definition."""
@@ -54,6 +56,11 @@ class ProcessingStack(core.Stack):
         creation_tasks["check_files_checksums"] = {
             "type": "batch",
             "parallel": True,
+            "result_path": aws_stepfunctions.JsonPath.DISCARD,
+        }
+        creation_tasks["check_stac_metadata"] = {
+            "type": "batch",
+            "parallel": False,
             "result_path": aws_stepfunctions.JsonPath.DISCARD,
         }
 
@@ -196,7 +203,7 @@ class ProcessingStack(core.Stack):
 
                 job_definition = aws_batch.JobDefinition(
                     self,
-                    f"{task_name}-job",
+                    f"{task_name}{JOB_DEFINITION_SUFFIX}",
                     container=aws_batch.JobDefinitionContainer(
                         image=aws_ecs.ContainerImage.from_asset(
                             directory=".",
@@ -237,7 +244,7 @@ class ProcessingStack(core.Stack):
                     step_tasks[task_name] = aws_stepfunctions_tasks.BatchSubmitJob(
                         self,
                         task_name,
-                        job_name=f"{task_name}-job",
+                        job_name=f"{task_name}{JOB_DEFINITION_SUFFIX}",
                         job_definition=job_definition,
                         job_queue=batch_job_queue,
                         array_size=aws_stepfunctions.JsonPath.number_at("$.content.iteration_size"),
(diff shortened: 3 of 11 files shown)


imincik:
@l0b0 , I see this PR is still in draft state, but I have one comment about meaning of "endpoint" in current DL design. According current design we have following endpoints which are relevant for this PR (think about them as REST API endpoints):

* datasets - already implemented
* datasets/{id}/versions/ - will be implemented in https://app.zenhub.com/workspaces/data-lake-5d3fb8ec6d87a44faafbe313/issues/linz/geospatial-data-lake-team/36. Dataset and dataset metadata validation, import processes are triggered by this endpoint, but they are not endpoints by them selves.

It would be best to avoid using term endpoint in this PR and move validation code out of `endpoints` dir (move it to `backend/processing` dir as suggested by #56. Thanks

.github/workflows/ci.yml
```diff
@@ -31,7 +32,7 @@ jobs:
               run: |
                   python -m pip install --upgrade pip
                   python -m pip install poetry
-                  python -m poetry install --extras='cdk datasets-endpoint'
+                  python -m poetry install --extras='cdk datasets-endpoint check-metadata-job'
```
imincik, line 35: * I suggest name of Poetry extra to be in format `<JOB-NAME>-job`

.github/workflows/ci.yml
```diff
@@ -81,7 +84,7 @@ jobs:
...
```
imincik, line 87: * I suggest name of Poetry extra to be in format `<JOB-NAME>-job`

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "datalake/backend/processing/check_metadata_content/stac-spec"]
+	path = datalake/backend/processing/check_metadata_content/stac-spec
```
imincik, line 2: * according Detailed Design, this job name (and directory name) should be `check-stac-metadata`.

datalake/backend/processing/check_metadata_content/task.py
```diff
@@ -0,0 +1,57 @@
...
+        bucket_name = parse_result.netloc
+        key = parse_result.path[1:]
+        response = client.get_object(Bucket=bucket_name, Key=key)
+        return response["Body"]
+
+    return read
+
+
+def main() -> int:
+    arguments = parse_arguments()
```
imincik, line 47: * according Detailed Design, this script should be called with following parameters
```
[
  "--dataset-id", "Ref::dataset_id",
  "--version-id", "Ref::version_id",
  "--type", "Ref::type",
  "--metadata-url", "Ref::metadata_url"
]
```

datalake/backend/processing/check_metadata_content/task.py
```diff
@@ -0,0 +1,57 @@
...
+    url_reader = s3_url_reader()
+
+    for url in arguments.urls:
+        validate_url(url, url_reader)
+
+    return 0
```
imincik, line 53: * according Detailed Design, this script should return following JSON

```
{
  "item": "<METADATA-FILE-NAME>",
  "success": <true|false>,
  "message": "<ERROR-MESSAGE>"
}
```

However,  I think we might like to replace `"item": "<METADATA-FILE-NAME>",` with `"item": "<METADATA-URL>",`

datalake/backend/processing/check_metadata_content/task.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
```
imincik, line 1: * I think it would be great to add at least some docstring documenting usage of this script

datalake/backend/processing/check_metadata_content/task.py
```diff
@@ -0,0 +1,57 @@
...
+import sys
```
imincik, line 2: Is this script following and checking all child metadata files ? See: https://confluence.linz.govt.nz/pages/viewpage.action?pageId=93566252

datalake/processing_stack.py
```diff
@@ -15,6 +15,9 @@
 )
 from aws_cdk.core import Tags
 
+CHECK_METADATA_CONTENT_JOB_DEFINITION_PREFIX = "check_metadata_content"
```
imincik, line 18: What's the purpose of using this constant `CHECK_METADATA_CONTENT_JOB_DEFINITION_PREFIX` instead of task name as it is used for all other tasks ?

datalake/processing_stack.py
```diff
@@ -56,6 +59,11 @@ def __init__(self, scope: core.Construct, stack_id: str, deploy_env, vpc, **kwar
             "parallel": True,
             "result_path": aws_stepfunctions.JsonPath.DISCARD,
         }
+        creation_tasks[CHECK_METADATA_CONTENT_JOB_DEFINITION_PREFIX] = {
+            "type": "batch",
+            "parallel": True,
```
imincik, line 64: This is not parallel task.

datalake/processing_stack.py
```diff
@@ -279,6 +287,7 @@ def __init__(self, scope: core.Construct, stack_id: str, deploy_env, vpc, **kwar
             step_tasks["check_flat_directory_structure"]
             .next(step_tasks["content_iterator"])
             .next(step_tasks["check_files_checksums"])
+            .next(step_tasks[CHECK_METADATA_CONTENT_JOB_DEFINITION_PREFIX])
```
imincik, line 290: This task should be first in State Machine workflow

tests/test_processing_stack.py
```diff
@@ -0,0 +1,20 @@
+from pytest import mark
+
+from datalake.processing_stack import (
+    CHECK_METADATA_CONTENT_JOB_DEFINITION_PREFIX,
+    JOB_DEFINITION_SUFFIX,
+)
+
+
+@mark.infrastructure
```
imincik, line 9: As we know, this is very light test.

imincik (changes requested):

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "datalake/backend/processing/check_metadata_content/stac-spec"]
```
imincik, line 1: Deployment fails on 

```
COPY failed: stat /var/lib/docker/tmp/docker-builder008079860/datalake/backend/processing/check_metadata_content/stac-spec/collection-spec/json-schema/collection.json: no such file or directory
[100%] fail: docker build --tag cdkasset-105dc3b20ec783a2e00ff6b5d249f6a7a9d65e41212c4518c36ceb5b14a5f419 --file datalake/backend/processing/check_metadata_content/Dockerfile . exited with error code 1: COPY failed: stat /var/lib/docker/tmp/docker-builder008079860/datalake/backend/processing/check_metadata_content/stac-spec/collection-spec/json-schema/collection.json: no such file or directory
```

We need to add something like below to our `reset-dev-env.bash` script

```
git submodule init 
git submodule update
```

.gitmodules
l0b0: I'll have to change that to `check_stac_metadata` to be able to import from it from Python.

datalake/backend/processing/check_metadata_content/task.py
l0b0: Not yet, no.

datalake/processing_stack.py
l0b0: It was necessary to make the name available to a test.

datalake/backend/processing/check_metadata_content/task.py
l0b0: Only the metadata URL is relevant for this job, so I don't see why we should. We might end up having to do that in some future iteration, of course.

datalake/backend/processing/check_metadata_content/task.py
l0b0: We know which metadata URL the script was called with. Do we really also need to return it?

datalake/backend/processing/check_metadata_content/task.py
l0b0: The script is self-documenting:

```
$ datalake/backend/processing/check_stac_metadata/task.py
usage: task.py [-h] --metadata-url METADATA_URL
task.py: error: the following arguments are required: --metadata-url
```

datalake/processing_stack.py
l0b0: Removed test and inlined constant as per our discussion.

.github/workflows/ci.yml
```diff
@@ -81,7 +85,8 @@ jobs:
               run: |
                   python -m pip install --upgrade pip
                   python -m pip install poetry
-                  python -m poetry install --extras='cdk datasets-endpoint'
+                  python -m poetry install \
+                      --extras='cdk datasets-endpoint check-metadata-content-job'
```
imincik, line 89: I found one thing - name of Poetry extra is not matching directory name of batch job
11:49 PM
check-metadata-content-job should be check-stac-metadata-job

imincik (approved):
Thanks Victor

PR #97, opened 2020-12-03
test: Randomise test data

l0b0:
Should be ready once https://github.com/linz/geospatial-data-lake/pull/99 is merged.


imincik (approved):
I like it, thanks

PR #99, opened 2020-12-04
Create dataset instances per test

l0b0:


imincik (approved):
Thanks. Looks as good idea to me.

PR #100, opened 2020-12-06
refactor: Pull out common dataset types list

l0b0:


imincik (approved):
Good, I like it. Thanks

PR #103, opened 2020-12-07
refactor: Simplify processing stack

l0b0:


imincik (commented):
@l0b0 , we don't have infrastructure test for this stack at all, so it would be very hard to say if this PR breaks something or not. For example I spent many iterations until I've got `batch_launch_template_data` in correct format (it is very fragile, any change of indentation was breaking it)

l0b0:
> @l0b0 , we don't have infrastructure test for this stack at all, so it would be very hard to say if this PR breaks something or not.

Good point, let's look into this later.

imincik:
@l0b0 , I need a chat with you how we want to progress with such infrastructure tests. Running whole processing work can take even 10-20 minutes, but I don't know about any other way how to do it.

imincik (changes requested):
Infrastructure tests are needed before such changes can be performed.

imincik:
@l0b0 , why _create_job_definition is not used in the same way as _create_lambda_invoke in
        content_iterator_lambda_invocation = self._create_lambda_invoke(
            "content_iterator", "$.content", [assets_table.grant_read_data]
        )
?
10:13 PM
This is one think which is confusing me and is quite hard to follow
10:14 PM
That loop which you replaced was doing approximately equal thing for Lambdas and Batches - package them and then create them
10:14 PM
So I guess it should be very similar in your PR as well
10:15 PM
Sorry for coming with more issues, but this looks to me as something which I would like to make clear
10:16 PM
I would rename those functions to _create_lambda_function and _create_batch_job
10:17 PM
and they should do only bundling

l0b0:
> @l0b0 , why `_create_job_definition` is not used in the same way as `_create_lambda_invoke` in
> `content_iterator_lambda_invocation = self._create_lambda_invoke( "content_iterator", "$.content", [assets_table.grant_read_data] )`
> ?
> This is one think which is confusing me and is quite hard to follow
> That loop which you replaced was doing approximately equal thing for Lambdas and Batches - package them and then create them
> So I guess it should be very similar in your PR as well
> Sorry for coming with more issues, but this looks to me as something which I would like to make clear
> I would rename those functions to _create_lambda_function and _create_batch_job
> and they should do only bundling. Do consistent task.

I agree. I have unpacked `_create_lambda_invoke` for simplicity.

imincik:
PR fails on test coverage now.

imincik:
@l0b0 , I just checked latest changes and at least after first couple of rounds of reading I can't say that pulling everything out to separate functions makes too much sense to me.

imincik:
To me, current code looks much harder to follow.

l0b0:
I couldn't follow the original code, where 50+ things were being created in the same context. For example, it was not obvious which variables were involved in creating any single one of the jobs. Now the init function mostly just delegates to create the top-level items to put in the state machine.

Since this is just a refactoring I'd be willing to reduce the coverage count to 83% to get this through.

imincik:
@l0b0 , I think my feedback to current state of this PR is that I am not against refactoring but I would ask you think if you can improve it to have some nicer methods which would allow us to create high level resources  - Lambdas, Batch Jobs, State Machine and DynamoDB table and will hide all details.

Thanks

l0b0:
> @l0b0 , I think my feedback to current state of this PR is that I am not against refactoring but I would ask you think if you can improve it to have some nicer methods which would allow us to create high level resources - Lambdas, Batch Jobs, State Machine and DynamoDB table and will hide all details.

@MitchellPaff mentioned how inheriting from `core.Construct` seems to be the common pattern for this, so I've created constructs corresponding to the methods we had before. This should be a lot less ad hoc, and hopefully more extensible.

imincik:
@l0b0 Best way how to debug latest tests failure is to deploy stacks to dev AWS account and run this test manually and then check error message from Batch job in CloudWatch.

MitchellPaff:
Should we also refactor the "datasets" table in the storage stack to use the new Table construct?

MitchellPaff (approved):

PR #106, opened 2020-12-07
ci: Wait for stack teardown before deploying

l0b0:
This should allow us to run multiple CI pipelines in parallel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,9 @@ jobs:
                   python -m pip install poetry
                   python -m poetry install --extras='cdk datasets-endpoint'
 
+            - name: Install AWS CLI
+              run: sudo apt install awscli
+
             - name: Use Node.js 12.x for CDK deployment
               uses: actions/setup-node@v2.1.2
               with:
@@ -120,6 +123,12 @@ jobs:
               run: |
                   DEPLOY_ENV=ci
                   echo "DEPLOY_ENV=$DEPLOY_ENV" | tee -a $GITHUB_ENV
+
+            - name: Wait for teardown of existing stacks
+              run: |
+                  timeout 2h \
+                      ./.github/workflows/wait-for-stack-teardown.bash \
+                      geospatial-data-lake-ci geospatial-data-lake-processing-ci
 
             - name: Deploy AWS stacks for testing
               run: |
--- a/.github/workflows/wait-for-stack-teardown.bash
+++ b/.github/workflows/wait-for-stack-teardown.bash
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset
+
+for stack_name
+do
+    while true
+    do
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -121,6 +124,10 @@ jobs:
                   DEPLOY_ENV=ci
                   echo "DEPLOY_ENV=$DEPLOY_ENV" | tee -a $GITHUB_ENV
 
+            - name: Wait if another CI AWS already stack exists
+              run: |
+                  timeout 2h ./wait-for-stack-teardown.bash geospatial-data-lake-ci
```
imincik, line 129: @l0b0 , can you add check for `geospatial-data-lake-networking-ci` and `geospatial-data-lake-processing-ci` stacks once `wait-for-stack-teardown.bash` can support multiple stacks ?

wait-for-stack-teardown.bash
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
imincik, line 1: What about some better place for this script than root src directory ? Put it in to `utils` dir ?

l0b0: I've put it with the workflow files, where it's being called from.

.github/workflows/ci.yml
l0b0: It already supports multiple stacks, just pass any number of stack names to the script and it'll wait for all of them to be torn down. But neither of those stacks are currently being torn down by CI, so I don't quite understand.

wait-for-stack-teardown.bash
imincik: strange, then GH is not showing full file name path to me

.github/workflows/ci.yml
imincik: I see it does support multiple stacks, that's why I am saying that it might be good to add `geospatial-data-lake-networking-ci` and `geospatial-data-lake-processing-ci`

l0b0: But if we add those two this task will always fail, because at least the `geospatial-data-lake-networking-ci` stack isn't torn down by CI.

imincik: Sorry, only `geospatial-data-lake-processing-ci` needs to be added.

l0b0: Done :)

imincik:
I think it doesn't work, I was checking AWS account and the only stack which is deployed is `networking` and our CI is polling:

```
Waiting for geospatial-data-lake-ci teardown…
135
Waiting for geospatial-data-lake-ci teardown…
136
Waiting for geospatial-data-lake-ci teardown…
137
Waiting for geospatial-data-lake-ci teardown…
138
Waiting for geospatial-data-lake-ci teardown…
139
Waiting for geospatial-data-lake-ci teardown…
140
Waiting for geospatial-data-lake-ci teardown…
141
Waiting for geospatial-data-lake-ci teardown…
```

imincik:
@l0b0 , I would change commit message to `ci: wait for AWS stacks teardown before deploying` (use ci:)

imincik (approved):
Looks working thanks.

imincik (approved):

PR #121, opened 2020-12-10
feat: Use Python 3.8

### LINKED ISSUE #109: Use Python 3.8 across the board
We've already encountered some issues with Python 3.6 (cattrs no longer supporting it and less support for type annotations), and we have no reason to support multiple Python versions in production.

To do:

- [x] Set Python version in `.python-version`.
- [ ] Use a single version in CI, removing the need for the strategy matrix.
- [ ] Use the `PYTHON_3_8` runtime for AWS Lambda jobs.
- [ ] Change the version in `pyproject.toml` to `^3.8,<3.9`.

l0b0:
Closes #109

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
         runs-on: ubuntu-20.04
         strategy:
             matrix:
-                python: ['3.6', '3.8']
+                python: ['3.8']
 
         steps:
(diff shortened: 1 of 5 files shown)


imincik (dismissed):
I like switching to 3.8, but when forcing it we should add pyenv installation and basic usage instructions.

imincik (approved):
I like this change, but I would prefer to merge #122 first.

PR #122, opened 2020-12-10
feat: Create dev env setup script & instructions

### LINKED ISSUE #123: Add script to set up development environment

l0b0:
Closes  #123.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,20 +88,21 @@ Re-run `. .venv/bin/activate` in each shell.
 
 
 ## AWS Infrastructure Deployment (CDK Stack)
-* Get AWS credentials (see: https://www.npmjs.com/package/aws-azure-login)
+1. Get AWS credentials (see: https://www.npmjs.com/package/aws-azure-login)
 
     ```bash
-    $ aws-azure-login -p <AWS-PROFILE-NAME>
+    aws-azure-login --profile=<AWS-PROFILE-NAME>
     ```
-* Deploy CDK stack
+1. Deploy CDK stack
 
     ```bash
-    $ cd infra
-    $ export ENVIRONMENT_TYPE=dev|nonprod|prod
-    $ cdk --profile <AWS-PROFILE-NAME> bootstrap aws://unknown-account/ap-southeast-2
-    $ cdk --profile <AWS-PROFILE-NAME> deploy --all
+    cd infra
+    export ENVIRONMENT_TYPE=dev|nonprod|prod
+    cdk --profile=<AWS-PROFILE-NAME> bootstrap aws://unknown-account/ap-southeast-2
+    cdk --profile=<AWS-PROFILE-NAME> deploy --all
     ```
 
+If you `export AWS_PROFILE=<AWS-PROFILE-NAME>` you won't need the `--profile=<AWS-PROFILE-NAME>` arguments above.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ Re-run `. .venv/bin/activate` in each shell.
 
     ```bash
     cd infra
-    export ENVIRONMENT_TYPE=dev|nonprod|prod
+    export DEPLOY_ENV=dev # Or 'ci', 'prod'
     cdk --profile=<AWS-PROFILE-NAME> bootstrap aws://unknown-account/ap-southeast-2
     cdk --profile=<AWS-PROFILE-NAME> deploy --all
     ```
--- a/README.md
+++ b/README.md
@@ -4,88 +4,106 @@
 Central storage, management and access for important geospatial datasets
 Developed by [Land Information New Zealand](https://github.com/linz)
 
+## Development setup
 
-## Dependencies Installation
-### Python Virtual Environment (for Python CLI and AWS CDK)
-* Create and activate a Python virtual environment
+One-time setup, assuming you are in the project directory:
 
-    ```bash
-    $ python3 -m venv .venv
-    $ source .venv/bin/activate
-    ```
-* Upgrade pip
+1. [Install and enable `pyenv`](https://github.com/pyenv/pyenv#installation):
+    1. Install Python build environment:
 
-    ```bash
-    $ pip install --upgrade pip
-    ```
-* [Install Poetry](https://python-poetry.org/docs/#installation)
-* Install the dependencies:
+        ```bash
+        sudo apt-get update
+        sudo apt-get install --no-install-recommends build-essential curl libbz2-dev libffi-dev liblzma-dev libncurses5-dev libreadline-dev libsqlite3-dev libssl-dev libxml2-dev libxmlsec1-dev llvm make tk-dev wget xz-utils zlib1g-dev
+        ```
+   1. `curl https://pyenv.run | bash`
+   1. Add the following to ~/.bashrc (wraps the upstream instructions to not do anything if `pyenv` is not installed):
 
-    ```bash
-    $ poetry install --extras='cdk datasets-endpoint'
-    ```
+       ```bash
+       # Pyenv <https://github.com/pyenv/pyenv>
+       if [[ -e "${HOME}/.pyenv" ]]
+       then
+           PATH="${HOME}/.pyenv/bin:${PATH}"
+           eval "$(pyenv init -)"
+           eval "$(pyenv virtualenv-init -)"
+       fi
+       ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4,50 +4,18 @@
...
-
-    ```bash
-    $ pip install --upgrade pip
-    ```
-* [Install Poetry](https://python-poetry.org/docs/#installation)
-* Install the dependencies:
-
-    ```bash
-    $ poetry install --extras='cdk datasets-endpoint'
-    ```
```
imincik, line 26: I would prefer copy/paste instructions for Ubuntu rather then those URL to installation instructions which are not always very clear. For example in case of Pyenv installation it was not very helpful to me. I rather went to `https://github.com/pyenv/pyenv-installer` and did `$ curl https://pyenv.run | bash` and then `$ pyenv install`. So for me having those commands in our instructions would be much better.

README.md
```diff
@@ -4,50 +4,18 @@
...
+1. [Install and enable `pyenv`](https://github.com/pyenv/pyenv#installation).
+   - Verify by running `diff <(python <<< 'import platform; print(platform.python_version())') .python-version`.
+1. [Install Poetry](https://python-poetry.org/docs/#installation).
+1. [Install and enable `nvm`](https://github.com/nvm-sh/nvm#installing-and-updating).
+1. [Install latest `npm` LTS](https://github.com/nvm-sh/nvm#long-term-support).
+1. Run `./reset-dev-env.bash` to install packages.
 
-### AWS CDK Environment (AWS Infrastructure)
```
imincik, line 28: I would prefer keeping those copy/paste commands to having URLs to installation  instructions. If you want to keep those URLs it is OK, but just as additional information to copy/paste commands.

imincik (changes requested):
@l0b0 , I like this idea,but left some comments

README.md
imincik: Also it would be great to have Python build deps installation instructions

```
$ sudo apt-get update; sudo apt-get install --no-install-recommends make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev
```

reset-dev-env.bash
```diff
@@ -0,0 +1,15 @@
...
+set -o errexit
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+npm install
+
+poetry install --extras='cdk datasets-endpoint'
+
+# shellcheck source=/dev/null
+. .venv/bin/activate
```
imincik, line 12: Why not using `pyenv activate` ?

reset-dev-env.bash
```diff
@@ -0,0 +1,15 @@
...
+
+pre-commit install --hook-type=commit-msg --overwrite
+pre-commit install --hook-type=pre-commit --overwrite
```
imincik, line 15: It would be great if this script would add `cdk` executable to  `PATH`

README.md
imincik: This looks like as a nice instructions https://realpython.com/intro-to-pyenv/#why-use-pyenv

imincik (commented):
.

reset-dev-env.bash
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+npm install
+
+poetry install --extras='cdk datasets-endpoint'
```
imincik, line 9: Why `poetry install` is executed before Python venv was activated ?

README.md
l0b0: The problem with copying the instructions is that they might change upstream at any point. Would it be better to suggest an improvement to the documentation upstream maybe?

README.md
l0b0: Having *both* the links and the instructions is a good compromise. I'll do that. Thanks!

reset-dev-env.bash
l0b0: I didn't know that was a thing. I'll look into it.

reset-dev-env.bash
l0b0: That won't propagate to the parent shell. But we should include instructions for doing that.

reset-dev-env.bash
l0b0: That's how Poetry works - it bootstraps the virtualenv before installing the software.

reset-dev-env.bash
l0b0: It looks like [the subproject](https://github.com/pyenv/pyenv-virtualenv/issues) is dead, and besides the command doesn't work:

```
$ pyenv activate 
pyenv-virtualenv: version `3.8.6' is not a virtualenv
```

README.md
imincik: Thanks

imincik: I tested process and works ! I'll have few small comments in new thread.

reset-dev-env.bash
imincik: OK, thanks

imincik: Script tested and works.

reset-dev-env.bash
imincik: Thanks. I had the same message and was wondering if I am doing something wrong.

README.md
```diff
@@ -4,88 +4,107 @@
...
+
+Re-run `. .venv/bin/activate` in each shell.
 
 
 ## AWS Infrastructure Deployment (CDK Stack)
+* Set AWS profile name: `export AWS_PROFILE=<AWS-PROFILE-NAME>`
 * Get AWS credentials (see: https://www.npmjs.com/package/aws-azure-login)
 
     ```bash
-    $ ./node_modules/.bin/aws-azure-login -p <AWS-PROFILE-NAME>
```
imincik, line 57: I think you must specify 'AWS-PROFILE-NAME` otherwise it will probably overwrite your default profile which can create mess in your AWS credentials and can be even dangerous.

imincik (changes requested):
@l0b0 , I left one last comment.

README.md
l0b0: Can you be a bit more specific? What is it going to overwrite, and in which circumstances? It works for me, and doesn't change `$AWS_PROFILE`:

```
$ echo $AWS_PROFILE 
li-datalake-nonprod
$ aws-azure-login
Logging in with profile 'li-datalake-nonprod'...
Using AWS SAML endpoint https://signin.aws.amazon.com/saml
? Session Duration Hours (up to 12): 12
Assuming role arn:aws:iam::632223577832:role/AccountAdminRole
$ echo $AWS_PROFILE 
li-datalake-nonprod
```

imincik: Using `-p` parameter you specify which AWS profile in `~/.aws/config` and `~/.aws/credentials` you want to configure. And if it works like I expect it should, by not specifying any profile `-p` you are configuring your default profile in your `~/.aws` configuration.

imincik: I my case, I have heaps of AWS profiles in my config files.

l0b0: Per [the documentation](https://github.com/sportradar/aws-azure-login#logging-in) and my experience (I haven't set a default profile in `~/.aws/config`) it uses the value of `$AWS_PROFILE`. This is a common pattern: the configuration file has the lowest priority, followed by environment variables, followed by the command line parameter. By specifying the environment variable I don't have to specify the profile to `aws-azure-login` *or* `cdk`, which makes things a bit easier.

imincik: Yes, if you use `AWS_PROFILE` variable then it is the same as setting `-p`. I always set `-p`, don't use `AWS_PROFILE` variable. I think using `--profile` or `-p` with CLI is more common. But anyway, one of those ways to say which profile to use has to be mentioned.

l0b0: The first entry in the list is

> Set AWS profile name: `export AWS_PROFILE=<AWS-PROFILE-NAME>`

Is that not what you mean?

imincik: Yes, I think it is more common to use `--profile` parameter then `AWS_PROFILE` env variable. I would use `--profile`.

l0b0: Cheers, done.

imincik (approved):
Thanks @l0b0

PR #124, opened 2020-12-10
refactor: Reduce max code complexity

l0b0:


imincik (approved):
Good change. Thanks

imincik (commented):
@l0b0 , do we need to add `pylint-mccabe` to Poetry python dependencies ?

l0b0:
> @l0b0 , do we need to add `pylint-mccabe` to Poetry python dependencies ?

No, it checks the complexity without installing that. Based on the [release history](https://pypi.org/project/pylint-mccabe/#history) I think that's an abandoned separate extension. Strange that it hasn't been marked as such.

PR #126, opened 2020-12-11
chore: Ignore CDK generated context file

l0b0:


imincik (approved):
Thanks for this @l0b0

PR #127, opened 2020-12-11
test: Pull boto client creation into functions

l0b0:


imincik:
What's the advantage of this change ?

l0b0:
> What's the advantage of this change ?

It's in the commit message:

Fixes the issue where commands like `pytest -m 'not infrastructure'
infra/tests` fails because it tries to create a boto3 client when
looking for test markers:

> botocore.exceptions.NoRegionError: You must specify a region.

imincik (approved):
Thanks @l0b0

PR #130, opened 2020-12-13
docs: Start a code review checklist

l0b0:


imincik (commented):
* At least in my browser it looks like it is file name containing spaces ? I would try to avoid spaces in file names if that's true

* I apologize, but it is too long and contains too much details (more like a blog post - I did read you blog post about this topic). I don't think we would be able to keep this standard with all other topics

l0b0:
>     * At least in my browser it looks like it is file name containing spaces ? I would try to avoid spaces in file names if that's true

It's a markdown file which is not included anywhere, so it doesn't matter if there are spaces in the filename.

>     * I apologize, but it is too long and contains too much details (more like a blog post - I did read you blog post about this topic). I don't think we would be able to keep this standard with all other topics

That's fair, but do we have anywhere more convenient to put this? Content in Confluence isn't peer reviewed by default, and isn't public, so it wouldn't be available to externals. How about putting it in the project wiki?

imincik:
Wiki is OK to me.

PR #132, opened 2020-12-14
refactor: Pull CDK app into main directory

### LINKED ISSUE #94: Avoid having to disable import errors in infrastructure code
CDK seems to create a directory layout which breaks IDE integration and Pylint validation. Just [using relative imports does not work](https://github.com/linz/geospatial-data-lake/pull/95/commits/79fb2273cd1cac2ea9f82b17922794967e149c88).

_Originally posted by @imincik in https://github.com/linz/geospatial-data-lake/pull/56#discussion_r535303416_

l0b0:
Closes #94.


imincik:
@l0b0 , this change might make a sense, but to me it looks to me that some more directories structure refactoring might be needed (for example moving  `backend` dir to `datalake` ?).

l0b0:
> @l0b0 , this change might make a sense, but to me it looks to me that some more directories structure refactoring might be needed (for example moving `backend` dir to `datalake` ?).

Would that maybe fit better into another PR?

imincik:
We can do it in next PR, but really should be done then.

imincik (dismissed):
Thanks

imincik (approved):
I am ok with that. I trust your intuition that this new layout will have advantages over current one.

PR #135, opened 2020-12-15
ci: use CI AWS account for CI

imincik:
Use CI AWS account for our CI jobs


imincik:
@l0b0 , as exception, I'll merge this PR without review.

PR #138, opened 2020-12-16
feat: Put backend within datalake

l0b0:
Based on https://github.com/linz/geospatial-data-lake/pull/132


imincik (approved):
Thanks Victor

Issue #134, opened 2020-12-15
Enable deploying test instance without conflicting with CI instance


l0b0:
Currently we have five elastic IPs, and each networking stack uses three. This means that if we deploy a test instance we break all CI pipelines, because they are unable to set up a networking stack. If we work around this by deploying a CI instance for testing purposes then any other CI pipelines will fail after waiting for two hours for the teardown of the existing stacks, which is not happening.

To resolve this we have a few options:

- Use fewer EIPs per networking stack. We should be able to use dynamic IPs for everything or almost everything.
- Pay for more EIPs. Probably not a good idea in the long run.
- Use a separate account for test instances. Probably not feasible, and also probably more expensive.

imincik:
Good option is to use single AZ (with public IP address) if `DEPLOY_ENV` is not `prod`

PR #142, opened 2020-12-17
feat(infra): allow usage of existing VPC coming with AWS account

imincik:
This change is introducing following features:

* possibility to use existing VPC provided by/coming with AWS account
* or in case we create VPC by our networking stack, it is created with limited networking resources (NAT GW, EIPs) if deployment environment is not `prod`. This will reduce cost of networking resources for test and CI deployments and will
reduce usage of EIPs (which is limited to 5 EIPs/AWS account).

Closes: #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,10 +93,16 @@ Re-run `. .venv/bin/activate` in each shell.
     ```bash
     aws-azure-login --profile=<AWS-PROFILE-NAME>
     ```
+1. Environment variables
+* **DEPLOY_ENV:** set deployment environment. Recommended values: prod, nonprod, ci, dev or any
+    string without spaces. Default: dev.
+* **DATALAKE_USE_EXISTING_VPC:** determine if networking stack will use existing VPC in target AWS
+    account or will create a new one. Existing VPC must to be used must contain following tags -
+    "ApplicationName": "geospatial-data-lake", "ApplicationLayer": "networking".
+    Allowed values: true, false. Default: false.
 1. Deploy CDK stack
 
     ```bash
-    export DEPLOY_ENV=dev # Or 'ci', 'prod'
     cdk --profile=<AWS-PROFILE-NAME> bootstrap aws://unknown-account/ap-southeast-2
     cdk --profile=<AWS-PROFILE-NAME> deploy --all
     ```
--- a/app.py
+++ b/app.py
@@ -23,6 +23,7 @@ networking = NetworkingStack(
     "networking",
     stack_name=f"geospatial-data-lake-networking-{ENV}",
     env={"region": os.environ["CDK_DEFAULT_REGION"], "account": os.environ["CDK_DEFAULT_ACCOUNT"]},
+    deploy_env=ENV,
 )
 
 storage = StorageStack(
--- a/app.py
+++ b/app.py
@@ -24,6 +24,7 @@ networking = NetworkingStack(
     stack_name=f"geospatial-data-lake-networking-{ENV}",
     env={"region": os.environ["CDK_DEFAULT_REGION"], "account": os.environ["CDK_DEFAULT_ACCOUNT"]},
     deploy_env=ENV,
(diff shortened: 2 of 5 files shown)


imincik:
This PR might need manual dropping of CI networking stack to succeed.

datalake/networking_stack.py
```diff
@@ -32,5 +32,6 @@ def __init__(self, scope: core.Construct, stack_id: str, **kwargs) -> None:
                     reserved=True,
                 ),
             ],
+            max_azs=99 if deploy_env == "prod" else 1,
```
l0b0, line 35: We should pull out an [enum](https://docs.python.org/3/library/enum.html) for the environment names sooner rather than later. Up to you whether you want to create a separate PR for that.

l0b0:
> This PR might need manual dropping of CI networking stack to succeed.

Why?

datalake/networking_stack.py
imincik: I would rather add enum in other PR. It needs to be used at multiple places.

imincik: > Also, I expect max_azs=None is more appropriate than 99.

From CDK docs describing `max_azs` parameter looks like using value such as `99` is recommended.

"""
To use “all AZs” available to your account, use a high number (such as 99).
"""

imincik: Docs is not saying what's going to happen if I set None.

imincik:
@l0b0 , this PR was significantly changed.

app.py
```diff
@@ -23,6 +23,8 @@
     "networking",
     stack_name=f"geospatial-data-lake-networking-{ENV}",
     env={"region": os.environ["CDK_DEFAULT_REGION"], "account": os.environ["CDK_DEFAULT_ACCOUNT"]},
+    deploy_env=ENV,
+    use_existing_vpc=True,  # use already existing VPC, pre-configured in AWS account
```
l0b0, line 27: As the code stands it's not possible to call this with `False` without changing the code. I'd probably do something like `use_existing_vpc = str2bool(os.environ.get("USE_EXISTING_VPC"))` where

```
def str2bool(value: str) -> bool:
    if value == "True":
        return True
    if value == "False":
        return False
    raise ValueError(f"Not a valid boolean: '{value}'")
```

Up to you if you want to do that in a separate PR.

imincik: Yes , this is something which I am aware of. I was thinking about env variable as well.

imincik: Should be done in d1e2abaeab5a4c9946386678b3b06a97a246ffbe

README.md
```diff
@@ -93,10 +93,16 @@ Re-run `. .venv/bin/activate` in each shell.
     ```bash
     aws-azure-login --profile=<AWS-PROFILE-NAME>
     ```
+1. Environment variables
+* **DEPLOY_ENV:** set deployment environment. Recommended values: prod, nonprod, ci, dev or any
+    string without spaces
+* **DATALAKE_USE_EXISTING_VPC:** determine if networking stack will use existing VPC in target AWS
```
l0b0, line 99: We should document the default here.

app.py
```diff
@@ -16,13 +16,23 @@
...
+
 app = core.App()
 
 networking = NetworkingStack(
     app,
     "networking",
     stack_name=f"geospatial-data-lake-networking-{ENV}",
     env={"region": os.environ["CDK_DEFAULT_REGION"], "account": os.environ["CDK_DEFAULT_ACCOUNT"]},
+    deploy_env=ENV,
+    use_existing_vpc=str2bool(os.environ.get("DATALAKE_USE_EXISTING_VPC", "true")),
```
l0b0, line 35: Do we want to use `True` as the default? I would've thought `False` would be safer.

l0b0 (changes requested):

README.md
imincik: It is documented now and `DATALAKE_USE_EXISTING_VPC` default is `false`

l0b0 (approved):

PR #147, opened 2020-12-20
ci: Run tests in parallel with linter

### LINKED ISSUE #146: Run CI jobs in parallel
If the lint job fails it's still useful to see the result of the test job, and our CI would be faster if they ran in parallel. We should get rid of `needs: [linter]` in the `test` job.

As for the deploy job, we shouldn't make that parallel until we can reliably run an arbitrary number of them in parallel without clashing.

l0b0:
Closes #146.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,7 +52,6 @@ jobs:
 
 
     test:
-        needs: [linter]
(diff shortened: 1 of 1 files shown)


imincik (dismissed):
Thanks for this

imincik (approved):

Issue #148, opened 2020-12-21
Fix commit message checks


l0b0:
The "Check all commit messages in Pull Request" step does not seem to run in PRs ([example](https://github.com/linz/geospatial-data-lake/pull/147/checks?check_run_id=1586109420)). Instead the "Check last commit message" check runs, and that should only run after merging to master, to verify the merge commit.

I've [asked](https://github.community/t/how-to-set-a-job-to-run-when-pushing-to-a-pr-branch/151198) the community about this, because I can't figure out how to do this.

PR #150, opened 2020-12-21
feat: Split up deploy and test pipelines

l0b0:
Closes issue #148, as seen in https://github.com/linz/geospatial-data-lake/runs/1586803853?check_suite_focus=true#step:6:1


imincik (approved):
Looks good to me.

PR #151, opened 2020-12-21
fix(infra): run 'check_stac_metadata' test first in dataset validation

imincik:
`check_stac_metadata` step should run as first one.


l0b0 (commented):
Did you mean to commit just an empty file?

imincik:
Ah, thanks. I was wondering why I still see this this change in my diff.

imincik:
Fixed now.

l0b0 (approved):

PR #155, opened 2020-12-21
test(infra): add processing stack tests

### LINKED ISSUE #129: Infra: implement tests for processing stack

imincik:
closes #129

### DIFF
--- a/datalake/backend/tests/conftest.py
+++ b/datalake/backend/tests/conftest.py
@@ -4,12 +4,23 @@ Pytest configuration file.
 
 import logging
 
+import boto3
 import pytest
 
 from ..endpoints.datasets.model import DatasetModel
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
+
+
+@pytest.fixture()
+def batch_client():
+    return boto3.client("batch")
+
+
+@pytest.fixture()
+def stepfunctions_client():
+    return boto3.client("stepfunctions")
 
 
 @pytest.fixture()
--- a/datalake/backend/tests/test_processing_stack.py
+++ b/datalake/backend/tests/test_processing_stack.py
@@ -1,0 +1,158 @@
+import json
+import logging
+import os
+import time
+from io import BytesIO
+from json import dumps
+from typing import Any, Dict
+
+from pytest import mark
+
+from .utils import (
+    S3Object,
+    any_dataset_description,
+    any_dataset_id,
+    any_past_datetime_string,
+    any_safe_file_path,
+    any_safe_filename,
+)
+
+APPLICATION_TAG_KEY = "ApplicationName"
+APPLICATION_TAG_VAL = "geospatial-data-lake"
+
+ENVIRONMENT_TAG_KEY = "EnvironmentType"
+ENV = os.environ.get("DEPLOY_ENV", "dev")
+
+STAC_VERSION = "1.0.0-beta.2"
+
+MINIMAL_VALID_STAC_OBJECT: Dict[str, Any] = {
+    "stac_version": STAC_VERSION,
+    "id": any_dataset_id(),
+    "description": any_dataset_description(),
+    "links": [],
+    "license": "MIT",
+    "extent": {
+        "spatial": {"bbox": [[-180, -90, 180, 90]]},
+        "temporal": {"interval": [[any_past_datetime_string(), None]]},
+    },
+}
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class NoStateMachineFound(Exception):
+    pass
+
+
+class NoComputeEnvironmentFound(Exception):
+    pass
+
+
+def get_compute_environment(batch_client):
+    compute_environment_detection_response = batch_client.describe_compute_environments()
+
+    # make sure there are no more records in next pages
+    assert compute_environment_detection_response.get("nextToken") is None
+
+    for compute_environment in compute_environment_detection_response["computeEnvironments"]:
+        if (
+            compute_environment["tags"][ENVIRONMENT_TAG_KEY] == ENV
+            and compute_environment["tags"].get(APPLICATION_TAG_KEY, None) == APPLICATION_TAG_VAL
+        ):
+            datalake_compute_environment = compute_environment
+            logger.info("Datalake Compute Environment: %s", datalake_compute_environment)
+
+            return datalake_compute_environment
+
+    raise NoComputeEnvironmentFound(APPLICATION_TAG_VAL, ENV)
+
+
+def launch_compute_environment_resources(batch_client, datalake_compute_environment):
+    """
+    Trigger immediate launch of computing resources. This will significantly speed up Batch jobs
+    execution.
+    """
+    if (
+        int(datalake_compute_environment["computeResources"]["minvCpus"]) < 1
+        or int(datalake_compute_environment["computeResources"]["desiredvCpus"]) < 1
+    ):
+        batch_client.update_compute_environment(
+            computeEnvironment=datalake_compute_environment["computeEnvironmentName"],
+            computeResources={"minvCpus": 1, "desiredvCpus": 1},
+        )
+
+
+def get_state_machine(stepfunctions_client):
+    state_machines_detection_response = stepfunctions_client.list_state_machines()
+
+    # make sure there are no more records in next pages
+    assert state_machines_detection_response.get("nextToken") is None
+
+    for state_machine in state_machines_detection_response["stateMachines"]:
+
+        tags_detection_response = stepfunctions_client.list_tags_for_resource(
+            resourceArn=state_machine["stateMachineArn"]
+        )
+
+        # reformat returned array of tags to dictionary
+        state_machine_tags = {tag["key"]: tag["value"] for tag in tags_detection_response["tags"]}
+
+        if (
+            state_machine_tags[ENVIRONMENT_TAG_KEY] == ENV
+            and state_machine_tags.get(APPLICATION_TAG_KEY, None) == APPLICATION_TAG_VAL
+        ):
+            datalake_state_machine = state_machine
+            logger.info("Datalake State Machine: %s", datalake_state_machine)
+
+            return datalake_state_machine
+
+    raise NoStateMachineFound(APPLICATION_TAG_VAL, ENV)
+
+
+@mark.timeout(1200)
+@mark.infrastructure
+def test_should_successfully_run_dataset_version_creation_process(
+    batch_client, stepfunctions_client
+):
+
+    metadata_file = "{}/{}.json".format(any_safe_file_path(), any_safe_filename())
+    metadata_content = dumps(MINIMAL_VALID_STAC_OBJECT)
+    s3_bucket = f"linz-geospatial-data-lake-dataset-source-{ENV}"
+
+    with S3Object(
+        BytesIO(f"{metadata_content}".encode()),
+        f"{s3_bucket}",
+        f"{metadata_file}",
+    ) as s3_metadata_file:
+
+        state_machine_input = json.dumps(
+            {
+                "dataset_id": "1xyz",
+                "version_id": "2xyz",
+                "type": "RASTER",
+                "metadata_url": f"s3://{s3_metadata_file.bucket_name}/{s3_metadata_file.key}",
+            }
+        )
+
+        # speed-up Compute Environment launch
+        datalake_compute_environment = get_compute_environment(batch_client)
+        launch_compute_environment_resources(batch_client, datalake_compute_environment)
+
+        # launch State Machine
+        datalake_state_machine = get_state_machine(stepfunctions_client)
+        execution_response = stepfunctions_client.start_execution(
+            stateMachineArn=datalake_state_machine["stateMachineArn"], input=state_machine_input
+        )
+        logger.info("Executed State Machine: %s", execution_response)
+
+        # poll for State Machine State
+        while (
+            execution := stepfunctions_client.describe_execution(
+                executionArn=execution_response["executionArn"]
+            )
+        )["status"] == "RUNNING":
+            logger.info("Polling for State Machine state %s", "." * 6)
+            time.sleep(5)
+
+        assert execution["status"] == "SUCCEEDED", execution
--- a/app.py
+++ b/app.py
@@ -9,6 +9,7 @@ import os
 from aws_cdk import core
 
 from datalake.api_stack import APIStack
+from datalake.dataset_source_stack import DatasetSourceStack
 from datalake.networking_stack import NetworkingStack
 from datalake.processing_stack import ProcessingStack
 from datalake.storage_stack import StorageStack
@@ -58,6 +59,14 @@ api = APIStack(
     stack_name=f"geospatial-data-lake-api-{ENV}",
     env={"region": os.environ["CDK_DEFAULT_REGION"], "account": os.environ["CDK_DEFAULT_ACCOUNT"]},
     datasets_table=storage.datasets_table,
+)
+
+api = DatasetSourceStack(
+    app,
+    "dataset-source",
+    stack_name=f"geospatial-data-lake-dataset-source-{ENV}",
+    env={"region": os.environ["CDK_DEFAULT_REGION"], "account": os.environ["CDK_DEFAULT_ACCOUNT"]},
+    deploy_env=ENV,
 )
 
 # tag all resources in stack
--- a/cdk.json
+++ b/cdk.json
@@ -2,6 +2,7 @@
   "app": "python3 app.py",
   "context": {
     "data-lake-storage-bucket-name": "linz-geospatial-data-lake",
+    "data-lake-dataset-source-bucket-name": "linz-geospatial-data-lake-dataset-source",
     "@aws-cdk/core:enableStackNameDuplicates": "true",
     "@aws-cdk:enableDiffNoFail": "true",
     "@aws-cdk/core:stackRelativeExports": "true"
--- a/datalake/backend/processing/check_stac_metadata/Dockerfile
+++ b/datalake/backend/processing/check_stac_metadata/Dockerfile
@@ -25,4 +25,5 @@ RUN apt-get update \
 COPY --from=build /opt/.venv /opt/.venv
 
 COPY datalake/backend/processing/check_stac_metadata/stac-spec/collection-spec/json-schema/collection.json /opt/stac-spec/collection-spec/json-schema/collection.json
+COPY datalake/backend/processing/check_stac_metadata/stac-spec/catalog-spec/json-schema/catalog.json /opt/stac-spec/catalog-spec/json-schema/catalog.json
 COPY datalake/backend/processing/check_stac_metadata/task.py /opt/
--- a/datalake/backend/processing/check_stac_metadata/Dockerfile
+++ b/datalake/backend/processing/check_stac_metadata/Dockerfile
@@ -19,7 +19,7 @@ ENTRYPOINT ["/opt/.venv/bin/python", "/opt/task.py"]
 
 ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update \
-    && apt-get install --assume-yes --no-install-recommends python3 \
+    && apt-get install --assume-yes --no-install-recommends ca-certificates python3 \
     && rm -rf /var/lib/apt/lists/*
 
 COPY --from=build /opt/.venv /opt/.venv
--- a/datalake/backend/processing/content_iterator/task.py
+++ b/datalake/backend/processing/content_iterator/task.py
@@ -1,8 +1,8 @@
 def lambda_handler(event, _context):
     """Main Lambda entry point."""
 
-    total_size = 100
-    iteration_size = 10
+    total_size = 6
+    iteration_size = 2
 
     if "content" in event.keys():
         first_item = int(event["content"]["next_item"])
--- a/datalake/backend/tests/utils.py
+++ b/datalake/backend/tests/utils.py
@@ -2,13 +2,16 @@ import string
 from datetime import datetime, timedelta, timezone
 from random import choice, randrange
 from types import TracebackType
-from typing import Optional, Type
+from typing import BinaryIO, Optional, Type
 from uuid import uuid4
+
+import boto3
 
 from ..endpoints.datasets.common import DATASET_TYPES
 from ..endpoints.datasets.model import DatasetModel
 
 REFERENCE_DATETIME = datetime(2000, 1, 1, tzinfo=timezone.utc)
+DELETE_OBJECTS_MAX_KEYS = 1000
 
 
 def random_string(length: int) -> str:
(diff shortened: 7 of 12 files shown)


tests/test_processing_stack.py
```diff
@@ -0,0 +1,126 @@
...
+@mark.infrastructure
+def test_should_return_healthy_compute_environment(batch_client):
+    datalake_compute_environment = {}
+
+    response = batch_client.describe_compute_environments()
+    for compute_environment in response["computeEnvironments"]:
+        if (
+            APPLICATION_TAG_KEY in compute_environment["tags"].keys()
+            and compute_environment["tags"][APPLICATION_TAG_KEY] == APPLICATION_TAG_VAL
+        ):
```
l0b0, line 27: This could be simplified to `if compute_environment["tags"].get(APPLICATION_TAG_KEY, None) == APPLICATION_TAG_VAL:`.

tests/test_processing_stack.py
```diff
@@ -0,0 +1,126 @@
...
+
+    job_definitions_response = batch_client.describe_job_definitions()
+    for job_definition in job_definitions_response["jobDefinitions"]:
+        if (
+            APPLICATION_TAG_KEY in job_definition["tags"].keys()
+            and job_definition["tags"][APPLICATION_TAG_KEY] == APPLICATION_TAG_VAL
+        ):
+            datalake_job_definition = job_definition
+            logger.info("Job Definition: %s", datalake_job_definition)
+            break
```
l0b0, line 52: I would pull this out into a separate function. That way you can `raise NoJobDefinitionFound` or something if none of the job definitions match (avoiding a `KeyError` when running `datalake_job_definition["jobDefinitionName"]` below), and you won't need the `datalake_job_definition = {}` above.

tests/test_processing_stack.py
```diff
@@ -0,0 +1,126 @@
...
+    job_status = None
+    while job_status not in ("FAILED", "SUCCEEDED"):
+        job_status_response = batch_client.describe_jobs(
+            jobs=[
+                job_submit_response["jobId"],
+            ]
+        )
+        logger.info("Running Job: %s", job_status_response)
+        job_status = job_status_response["jobs"][0]["status"]
+        time.sleep(5)
```
l0b0, line 81: We have [pytest-timeout](https://pypi.org/project/pytest-timeout/) installed; we could `mark.timeout(…)` to avoid this job running forever.

Alternatively we could pull things apart a bit and use a generic waiter. Some untested code:

```
def wait_for(predicate: Callable[..., bool], timeout: timedelta, interval: timedelta, *args: Any, **kwargs: Any) -> None:
    end = datetime.now() + timeout
    while datetime.now() < end:
        if predicate(*args, **kwargs):
            return
        sleep(interval.seconds)

    assert False, "Timeout"
```

l0b0:
PS: I wrote an S3 file context manager for testing purposes, but I've managed to push the relevant tests closer to the unit level so I don't think it's in any of the PRs. Combined with `random_string` it should provide a nice way to avoid leaving data lying around after testing and to ensure tests are independent.

```
from types import TracebackType
from typing import BinaryIO, Optional, Type

import boto3

DELETE_OBJECTS_MAX_KEYS = 1000


class S3Object:
    def __init__(self, file_object: BinaryIO, bucket_name: str, key: str):
        self.file_object = file_object
        self.bucket_name = bucket_name
        self.key = key
        self.s3 = boto3.client("s3")

    def __enter__(self) -> "S3Object":
        self.s3.upload_fileobj(self.file_object, self.bucket_name, self.key)
        return self

    def __exit__(
        self,
        exc_type: Optional[Type[BaseException]],
        exc_val: Optional[BaseException],
        exc_tb: Optional[TracebackType],
    ) -> bool:
        version_list = self._get_object_versions()
        self._delete_object_versions(version_list)

        return False  # Propagate exception

    def _delete_object_versions(self, version_list):
        for index in range(0, len(version_list), DELETE_OBJECTS_MAX_KEYS):
            response = self.s3.delete_objects(
                Bucket=self.bucket_name,
                Delete={
                    "Objects": version_list[index : index + DELETE_OBJECTS_MAX_KEYS],
                },
            )
            print(response)

    def _get_object_versions(self):
        version_list = []
        object_versions_paginator = self.s3.get_paginator("list_object_versions")
        for object_versions_page in object_versions_paginator.paginate(Bucket=self.bucket_name):
            for marker in object_versions_page.get("DeleteMarkers", []):
                if marker["Key"] == self.key:
                    version_list.append({"Key": self.key, "VersionId": marker["VersionId"]})
            for version in object_versions_page.get("Versions", []):
                if version["Key"] == self.key:
                    version_list.append({"Key": self.key, "VersionId": version["VersionId"]})
        return version_list
```

Usage: `with S3Object(BytesIO(b"your content"), f"linz-geospatial-data-lake-{ENV}", filename) as s3_object:`

tests/test_processing_stack.py
imincik: Done in 24267c3eca463c99e4193102d1bf67c3c6854d92 . Thanks for this suggestion.

tests/test_processing_stack.py
imincik: Resources detection pulled out to functions in ec554077d939ab1f1e62e0ebdc5cf68a550c400d

tests/test_processing_stack.py
imincik: Done in e132cc941e96bbc1d50c24e3e208d6e2761090b3

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+            logger.info("Datalake Compute Environment: %s", datalake_compute_environment)
+            break
+
+    return datalake_compute_environment
+
+
+def launch_compute_environment_resources(batch_client, datalake_compute_environment):
+    """
+    Trigger immediate launch of computing resources. This will significantly speed up Batch jobs
+    execution.
```
l0b0, line 36: Nice!

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+        or int(datalake_compute_environment["computeResources"]["desiredvCpus"]) < 1
+    ):
+        batch_client.update_compute_environment(
+            computeEnvironment=datalake_compute_environment["computeEnvironmentName"],
+            computeResources={"minvCpus": 1, "desiredvCpus": 1},
+        )
+
+
+def detect_state_machine(stepfunctions_client):
+    state_machines_detection_response = stepfunctions_client.list_state_machines()
```
l0b0, line 49: `assert response["nextToken"] is None` like above.

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+    for state_machine in state_machines_detection_response["stateMachines"]:
+
+        tags_detection_response = stepfunctions_client.list_tags_for_resource(
+            resourceArn=state_machine["stateMachineArn"]
+        )
+
+        # reformat returned array of tags to dictionary
+        state_machine_tags = {}
+        for tag in tags_detection_response["tags"]:
+            state_machine_tags[tag["key"]] = tag["value"]
```
l0b0, line 61: This can be simplified to a `dict` comprehension

```
state_machine_tags = {tag["key"]: tag["value"] for tag in tags_detection_response["tags"]}
```

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+
+ENVIRONMENT_TAG_KEY = "EnvironmentType"
+ENVIRONMENT_TAG_VAL = os.environ.get("DEPLOY_ENV", "dev")
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def detect_compute_environment(batch_client):
+    compute_environment_detection_response = batch_client.describe_compute_environments()
```
l0b0, line 18: When using any AWS API calls which paginate we should (as discussed) do something like `assert response["nextToken"] is None` to get an alert if we ever need to start supporting pagination even in this test code.

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+        for tag in tags_detection_response["tags"]:
+            state_machine_tags[tag["key"]] = tag["value"]
+
+        if (
+            state_machine_tags[ENVIRONMENT_TAG_KEY] == ENVIRONMENT_TAG_VAL
+            and state_machine_tags.get(APPLICATION_TAG_KEY, None) == APPLICATION_TAG_VAL
+        ):
+            datalake_state_machine = state_machine
+            logger.info("Datalake State Machine: %s", datalake_state_machine)
+            break
```
l0b0, line 69: We could simplify the last three lines to just

```
            logger.info("Datalake State Machine: %s", state_machine)
            return state_machine
```

This has the added advantage that we won't ever return a non-existing variable.

We should also do something like `raise NoStateMachineFound(APPLICATION_TAG_VAL, ENVIRONMENT_TAG_VAL)` as the last line outside the loop to avoid returning `None` and kicking the ball further down the road.

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+    logger.info("Detected Compute Environments: %s", compute_environment_detection_response)
+
+    for compute_environment in compute_environment_detection_response["computeEnvironments"]:
+        if (
+            compute_environment["tags"][ENVIRONMENT_TAG_KEY] == ENVIRONMENT_TAG_VAL
+            and compute_environment["tags"].get(APPLICATION_TAG_KEY, None) == APPLICATION_TAG_VAL
+        ):
+            datalake_compute_environment = compute_environment
+            logger.info("Datalake Compute Environment: %s", datalake_compute_environment)
+            break
```
l0b0, line 28: This can be simplified in the same way as `detect_state_machine` below.

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+def test_should_successfully_run_dataset_version_creation_process(
+    batch_client, stepfunctions_client
+):
+
+    state_machine_input = """{
+        "dataset_id":"1xyz",
+        "version_id":"2xyz",
+        "type":"RASTER",
+        "metadata_url":"s3://linz-geospatial-data-lake-test-data/tasman_rural_2003_1m/collection.json"
+    }"""
```
l0b0, line 85: I would recommend using `dumps({"dataset_id": …})` to create a string out of a `dict` rather than hardcoding a string. That makes it easier to modify, and we won't have to worry about getting the formatting exactly right.

Also, it is not good to rely on any data maintained outside of the test. We should use `utils.Dataset` and [`S3Object`](https://github.com/linz/geospatial-data-lake/pull/179/commits/31a065e6603fe1714a92b2ec4a140cc40cdbe0b0) to create and tear down the information we need.

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+APPLICATION_TAG_VAL = "geospatial-data-lake"
+
+ENVIRONMENT_TAG_KEY = "EnvironmentType"
+ENVIRONMENT_TAG_VAL = os.environ.get("DEPLOY_ENV", "dev")
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def detect_compute_environment(batch_client):
```
l0b0, line 17: I feel like `get` would be more explicit than `detect` for these names. What do you think?

tests/test_processing_stack.py
```diff
@@ -0,0 +1,117 @@
...
+            state_machine_status_response["name"],
+            state_machine_status_response["status"],
+        )
+        logger.info("State Machine Input: %s", state_machine_status_response.get("input"))
+        logger.info("State Machine Output: %s", state_machine_status_response.get("output"))
+
+        machine_status = state_machine_status_response["status"]
+        time.sleep(5)
+
+    assert state_machine_status_response["status"] == "SUCCEEDED"
```
l0b0, line 117: Could we simplify the loop to maybe

```
while (execution := stepfunctions_client.describe_execution(executionArn=execution_response["executionArn"]))["status"] == "RUNNING":
    sleep(5)
assert execution["status"] == "SUCCEEDED", execution
```

? The logging is useful while building the test, but it's going to make the test runs much more verbose and a bit slower.

l0b0 (changes requested):

tests/test_processing_stack.py
imincik: Done in 9f529447d3faf75bb7c0ae62f9b923ae4cbce5c1

tests/test_processing_stack.py
imincik: Solved in 9f529447d3faf75bb7c0ae62f9b923ae4cbce5c1

tests/test_processing_stack.py
imincik: Done in 519d9039753ab1ad4eb1fbe52bad42533ca9a63b

tests/test_processing_stack.py
imincik: I get Pylint warning "inconsistent-return-statements" if I return from condition.

tests/test_processing_stack.py
imincik: Same as above

tests/test_processing_stack.py
imincik: JSON input done in b7b1d46619d9363b0ab138d485a81881332adaae

tests/test_processing_stack.py
imincik: Yes, I agree. Function names changed in a338c76dff1f809dc9c27b78fc3f38da7212af17

tests/test_processing_stack.py
imincik: Done in 29ca3fa32158cf6b4da381cfd5a7120ff541aae5 , thanks.

tests/test_processing_stack.py
l0b0: Did you change it to `raise` t the end of the function as well? Because then there's only one `return` statement.

datalake/backend/tests/test_processing_stack.py
```diff
@@ -0,0 +1,113 @@
...
+    return datalake_state_machine
+
+
+@mark.timeout(1200)
+@mark.infrastructure
+def test_should_successfully_run_dataset_version_creation_process(
+    batch_client, stepfunctions_client
+):
+
+    test_data_bucket = "s3://linz-geospatial-data-lake-test-data"
```
l0b0, line 83: This looks like reusable information; should it maybe be a constant in utils.py?

datalake/backend/tests/test_processing_stack.py
```diff
@@ -0,0 +1,113 @@
...
+    state_machine_input = json.dumps(
+        {
+            "dataset_id": "1xyz",
+            "version_id": "2xyz",
+            "type": "RASTER",
+            "metadata_url": f"{test_data_bucket}/tasman_rural_2003_1m/collection.json",
```
l0b0, line 89: If the code under test is actually going to have to read this file we should probably use an [`S3Object`](https://github.com/linz/geospatial-data-lake/pull/155#issuecomment-749300896) to populate it. That way we're not relying on external state, and the test can easily be tweaked to accommodate changes in the production code or infrastructure.

tests/test_processing_stack.py
imincik: Sorry, what I should change to `raise` ?

datalake/backend/tests/test_processing_stack.py
imincik: It is in separate variable only because otherwise I have too long line below (and Pylint wouldn't pass). And I guess this will need to be replaced by S3Object .

datalake/backend/tests/test_processing_stack.py
imincik: I agree, but I can't see your S3Object in source code right now. Do you want to add it ?

tests/test_processing_stack.py
l0b0: If we reach the end of the method without finding a match, then I would expect something like `raise NoStateMachineFound(APPLICATION_TAG_VAL, ENVIRONMENT_TAG_VAL)` as the last line.

datalake/backend/tests/test_processing_stack.py
l0b0: `TEST_DATA_BUCKET` shouldn't be any longer?

datalake/backend/tests/test_processing_stack.py
l0b0: You go ahead, I've got a copy on a branch which isn't merged yet.

datalake/backend/tests/conftest.py
```diff
@@ -12,6 +13,16 @@
 logger = logging.getLogger(__name__)
 
 
+@pytest.fixture()
+def batch_client():
+    return boto3.client("batch")
```
l0b0, line 18: I think this test might belong more in the root `tests` directory:

- It's an infrastructure test, so it requires the whole system to be deployed to succeed.
- It doesn't import anything from the datalake directory.

datalake/backend/tests/test_processing_stack.py
```diff
@@ -0,0 +1,113 @@
...
+            datalake_state_machine = state_machine
+            logger.info("Datalake State Machine: %s", datalake_state_machine)
+            break
+
+    return datalake_state_machine
+
+
+@mark.timeout(1200)
+@mark.infrastructure
+def test_should_successfully_run_dataset_version_creation_process(
```
l0b0, line 79: I think this test might belong more in the root `tests` directory:

- It's an infrastructure test, so it requires the whole system to be deployed to succeed.
- It doesn't import anything from the datalake directory.
- We already have pytest fixtures for batch in the top-level tests directory.

datalake/backend/tests/test_processing_stack.py
imincik: S3Object used in 8e466364d2d6f8b5d86e5be609d31174704e85d0

datalake/backend/tests/test_processing_stack.py
imincik: S3Object used in 8e466364d2d6f8b5d86e5be609d31174704e85d0

datalake/backend/tests/test_processing_stack.py
imincik: I would put all tests to `datalake/backend/tests`.

tests/test_processing_stack.py
imincik: @l0b0 , are you able to comment on this please ? Thanks

datalake/backend/tests/test_processing_stack.py
l0b0: Done in https://github.com/linz/geospatial-data-lake/pull/183

tests/test_processing_stack.py
imincik: Done in 7f60bdd1c351d8dab432c00ec6830644a6707b2e

tests/test_processing_stack.py
imincik: Done in 7f60bdd1c351d8dab432c00ec6830644a6707b2e

datalake/backend/tests/test_processing_stack.py
```diff
@@ -0,0 +1,133 @@
...
+        if (
+            compute_environment["tags"][ENVIRONMENT_TAG_KEY] == ENVIRONMENT_TAG_VAL
+            and compute_environment["tags"].get(APPLICATION_TAG_KEY, None) == APPLICATION_TAG_VAL
+        ):
+            datalake_compute_environment = compute_environment
+            logger.info("Datalake Compute Environment: %s", datalake_compute_environment)
+
+            return datalake_compute_environment
+
+    raise NoStateMachineFound(APPLICATION_TAG_VAL, ENVIRONMENT_TAG_VAL)
```
l0b0, line 46: I think you've swapped around `NoStateMachineFound` and `NoComputeEnvironmentFound` :)

datalake/backend/tests/test_processing_stack.py
```diff
@@ -0,0 +1,133 @@
...
+
+from .utils import S3Object, any_safe_file_path
+
+ENV = os.environ.get("DEPLOY_ENV", "dev")
+
+APPLICATION_TAG_KEY = "ApplicationName"
+APPLICATION_TAG_VAL = "geospatial-data-lake"
+
+ENVIRONMENT_TAG_KEY = "EnvironmentType"
+ENVIRONMENT_TAG_VAL = ENV
```
l0b0, line 16: This constant is just a copy of the other one, so one of them can be inlined.

datalake/backend/tests/collection.json
```diff
@@ -0,0 +1,104 @@
+{
+    "id": "test_dataset",
+    "stac_version": "1.0.0-beta.2",
+    "description": "This is dataset description.",
+    "links": [
+        {
+            "rel": "item",
+            "href": "s3://s3-bucket-name/datasets/xyz/1234.json",
```
l0b0, line 8: We'll be traversing these URLs as well, so they should also be `S3Objects`.

datalake/backend/tests/collection.json
```diff
@@ -0,0 +1,104 @@
...
+            ]
+        }
+    },
+    "license": "CC-BY-4.0",
+    "stac_extensions": [
+        "eo",
+        "proj",
+        "linz"
+    ],
+    "keywords": [
```
l0b0, line 54: We should make sure this dataset is as simple as humanly possible, to make sure the test and all its components are maintainable. Several of the attributes in this file, such as `keywords`, are not needed to create a valid STAC file. See for example `MINIMAL_VALID_STAC_OBJECT`.

datalake/backend/tests/collection.json
```diff
@@ -0,0 +1,104 @@
...
+    "description": "This is dataset description.",
+    "links": [
+        {
+            "rel": "item",
+            "href": "s3://s3-bucket-name/datasets/xyz/1234.json",
+            "type": "application/json"
+        },
+        {
+            "rel": "root",
+            "href": "s3://s3-bucket-name/catalog.json",
```
l0b0, line 13: `s3-bucket-name` is going to fail validation, so we probably want to use `f"linz-geospatial-data-lake-{ENV}"` here as well.

I would also use `any_safe_filename()` to clarify that the filename doesn't matter for the test.

l0b0 (changes requested):

datalake/backend/tests/test_processing_stack.py
```diff
@@ -0,0 +1,133 @@
...
+                "dataset_id": "1xyz",
+                "version_id": "2xyz",
+                "type": "RASTER",
+                "metadata_url": f"s3://{s3_object.bucket_name}/{s3_object.key}",
+            }
+        )
+
+        # speed-up Compute Environment launch
+        datalake_compute_environment = get_compute_environment(batch_client)
+        launch_compute_environment_resources(batch_client, datalake_compute_environment)
```
l0b0, line 115: Since this isn't strictly speaking *necessary* for the test it should probably be part of the `stepfunctions_client`.

datalake/backend/tests/test_processing_stack.py
```diff
@@ -0,0 +1,133 @@
...
+            computeEnvironment=datalake_compute_environment["computeEnvironmentName"],
+            computeResources={"minvCpus": 1, "desiredvCpus": 1},
+        )
+
+
+def get_state_machine(stepfunctions_client):
+    state_machines_detection_response = stepfunctions_client.list_state_machines()
+
+    # make sure there are no more records in next pages
+    assert state_machines_detection_response.get("nextToken", None) is None
```
l0b0, line 68: `.get("nextToken", None)` can be simplified to `.get("nextToken")`.

datalake/backend/tests/test_processing_stack.py
```diff
@@ -0,0 +1,133 @@
...
+        ):
+            datalake_compute_environment = compute_environment
+            logger.info("Datalake Compute Environment: %s", datalake_compute_environment)
+
+            return datalake_compute_environment
+
+    raise NoStateMachineFound(APPLICATION_TAG_VAL, ENVIRONMENT_TAG_VAL)
+
+
+def launch_compute_environment_resources(batch_client, datalake_compute_environment):
```
l0b0, line 49: How big is the difference between running this first and not running it? Comparing runs https://github.com/linz/geospatial-data-lake/runs/1718745340 and https://github.com/linz/geospatial-data-lake/runs/1718808531 it looks like a run without this takes 2m10s.

imincik (commented):
xxxx

datalake/backend/tests/test_processing_stack.py
imincik: done in 57de24f1545fd01666d5eec613b82066d8244122

datalake/backend/tests/collection.json
imincik: The problem with using `linz-geospatial-data-lake-{ENV}` S3 bucket is fact that this is Data Lake Storage bucket (destination). We should use some other bucket as source bucket.

imincik: Probably I should create new stack containing S3 bucket for input test datasets which we deploy only for testing

datalake/backend/tests/test_processing_stack.py
imincik: Done in eaaca6eaf34b053597b534f771e4d88f91d3de6d

datalake/backend/tests/test_processing_stack.py
imincik: You mean without running `launch_compute_environment_resources` function ? It can take 15 minutes  to launch compute resources without setting min and desired CPU to non zero value. Based on my experience it is very significant diference

datalake/backend/tests/test_processing_stack.py
imincik: I am not sure how much advantage or troubles it can cause. I would leave it as is for now.

datalake/backend/tests/collection.json
imincik: I used minimal STAC metadata file content now.

datalake/backend/tests/collection.json
imincik: In 3f81332ad4d2e2c7f3da46537381b1e04190a70d , I created dataset source stack containing separate S3 bucket for source datasets. We are going to need it anyway for customers which are not able to host source S3 bucket by them selves (this was discussed many times already).

datalake/backend/tests/collection.json
imincik: In 7c9a88f4c503a3bb403dab48a0d58a217af934f3 I used `MINIMAL_VALID_STAC_OBJECT` . I leave it on you to extend it when it will be needed.

l0b0 (dismissed):

l0b0:
Thank you!

l0b0 (approved):
Thank you!

PR #158, opened 2020-12-22
feat: Install only production dependencies in Docker container

l0b0:
Follow-up to https://github.com/linz/geospatial-data-lake/pull/154, since the original code was installing the development packages.


imincik (approved):
Good, thanks.

PR #159, opened 2020-12-22
fix: Remove deploy job dependencies

l0b0:


.github/workflows/deploy.yml
```diff
@@ -8,7 +8,6 @@ on:  # yamllint disable-line rule:truthy
 
 jobs:
     deploy:
-        needs: [linter, test]
```
imincik, line 11: I think linter and tests should pass before we start deployment otherwise we deploy some crap. Am I wrong ?

imincik (changes requested):

.github/workflows/deploy.yml
imincik: Actually it is OK. We run tests in PR and we don't allow merging without beeing updated with master

imincik (approved):

Issue #153, opened 2020-12-21
Implement "check_flat_directory_structure" check


imincik:
As discussed, make this part of the `check_stac_metadata` job:

- We're traversing the metadata files anyway, so there would be a lot of code duplication otherwise.
- If we had been planning to run `check_flat_directory_structure` and `check_stac_metadata` in parallel we could've done them both a bit faster, but with the current infrastructure we would be running them in serial. So rather than the ~100% overhead of running them in serial we can get a small overhead on `check_stac_metadata` to also check the flat structure.

l0b0:
Things to check:

- URLs which we don't traverse, that is, `rel=self` and `rel=root`.
- Should ignore repeated directory separators. That is, treat `foo/bar` and `foo//bar` as the same path.
- Only check the actual path. That is, in `https://s3.foo/path/root.json?foo=a/b&baz=c/d#e/f`, ignore everything except `/path/root.json`.
- Decode a single level of URL escaping, so that `foo%2Fbar` is treated like `foo/bar`, but `foo%252Fbar`(decoded to literally `foo%2Fbar`)  is not.

PR #167, opened 2021-01-06
feat: Validate flat directory structure

l0b0:
Closes issue #153.

### DIFF
--- a/datalake/backend/processing/check_flat_directory_structure/Dockerfile
+++ b/datalake/backend/processing/check_flat_directory_structure/Dockerfile
@@ -1,8 +1,0 @@
-FROM ubuntu:20.04
-ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get update \
-    && apt-get install --assume-yes --no-install-recommends python3 \
-    && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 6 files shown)


reset-dev-env.bash
```diff
@@ -8,7 +8,7 @@ git submodule update --init
 
 npm ci
 
-poetry install --extras='cdk check-stac-metadata-job datasets-endpoint'
```
imincik, line 11: This unrelated change. Better if goes in separate PR. Thanks

imincik (changes requested):
Left one comment, otherwise looks ok to me.

imincik:
@l0b0 , because this check is integrated to `check_stac_metadata` (not separate check as planned), would you please remove `check_flat_directory_structure` check from [processing stack](https://github.com/linz/geospatial-data-lake/blob/master/datalake/processing_stack.py#L50) and [here](https://github.com/linz/geospatial-data-lake/blob/master/datalake/processing_stack.py#L286) as well.

Also, [this directory](https://github.com/linz/geospatial-data-lake/tree/master/datalake/backend/processing/check_flat_directory_structure) needs to be deleted.

Thanks

imincik (approved):
Thanks @l0b0

imincik (approved):

PR #171, opened 2021-01-07
feat: Lint Dockerfiles

l0b0:


datalake/backend/processing/check_stac_metadata/Dockerfile
```diff
@@ -1,19 +1,26 @@
 FROM ubuntu:20.04 as build
 
 RUN apt-get update \
-    && apt-get install -y curl python3-pip \
+    && apt-get install --assume-yes --no-install-recommends curl=7.68.0-1ubuntu2.4 python3-pip=20.0.2-5ubuntu1.1 \
```
imincik, line 4: Apt package versions as above

datalake/backend/processing/check_files_checksums/Dockerfile
```diff
@@ -1,6 +1,8 @@
 FROM ubuntu:20.04
 ENV DEBIAN_FRONTEND noninteractive
-RUN apt update && apt install -y python3 && rm -rf /var/lib/apt/lists/*
+RUN apt-get update \
+    && apt-get install --assume-yes --no-install-recommends python3=3.8.2-0ubuntu2 \
```
imincik, line 4: @l0b0 , are you sure we need to specify apt packages versions here ? The only distribution upgrades are bug fixes. I think this is not necessary.

datalake/backend/processing/check_stac_metadata/Dockerfile
```diff
@@ -1,19 +1,26 @@
...
 
 
 FROM ubuntu:20.04
 
 ENTRYPOINT ["/opt/.venv/bin/python", "/opt/task.py"]
 
 ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get update && apt-get install -y python3 && rm -rf /var/lib/apt/lists/*
+RUN apt-get update \
+    && apt-get install --assume-yes --no-install-recommends python3=3.8.2-0ubuntu2 \
```
imincik, line 22: Apt package versions as above

imincik (changes requested):
@l0b0 , looks ok except one thing I asked to reconsider. Thanks

datalake/backend/processing/check_files_checksums/Dockerfile
l0b0: Pro locking versions:

- Makes it less likely to break things arbitrarily. Even OS package patches sometimes sneak in functional changes which impact systems. Reproducibility is a huge win.

Cons:

- Even if we install fixed versions of the top level packages, we might get updated versions of their dependencies. Apt AFAIK does not have a lockfile mechanism like NPM and Poetry, meaning that locking all the package dependencies is a bit more of a chore.
- [Dependabot does not yet patch Dockerfile package installs](https://github.com/dependabot/dependabot-core/issues/2129).

Once Dependabot supports `apt-get` within Dockerfiles that fixes both of the above - we could just lock the package dependencies in the `apt-get` command *once* and Dependabot would take care of keeping them up to date. So on balance I agree for now.

imincik (approved):
Thanks @l0b0

PR #173, opened 2021-01-08
feat(infra): add support to store and read data in S3 during processing

imincik:
see: https://confluence.linz.govt.nz/display/LGDL/Dataset+Version+Endpoint

### DIFF
--- a/datalake/processing_stack.py
+++ b/datalake/processing_stack.py
@@ -9,6 +9,7 @@ from aws_cdk import (
     aws_ecs,
     aws_iam,
     aws_lambda,
+    aws_s3,
     aws_stepfunctions,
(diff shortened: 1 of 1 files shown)


datalake/processing_stack.py
```diff
@@ -184,6 +199,7 @@ def __init__(self, scope: core.Construct, stack_id: str, deploy_env, vpc, **kwar
                         ),
                     ),
                 )
+                assets_bucket.grant_read_write(lambda_function)
```
l0b0, line 202: This gives write access to *all* the lambdas. We should only give write access to the content parser.

l0b0 (changes requested):

imincik:
This PR is replaced by #174

PR #174, opened 2021-01-11
feat(infra): add processing assets storage table

imincik:
See: https://confluence.linz.govt.nz/display/LGDL/Dataset+Version+Endpoint


datalake/processing_stack.py
```diff
@@ -21,12 +22,32 @@
...
     """Data Lake processing stack definition."""
 
-    # pylint: disable=too-many-locals
+    # pylint: disable=too-many-locals,too-complex
     def __init__(self, scope: core.Construct, stack_id: str, deploy_env, vpc, **kwargs) -> None:
         super().__init__(scope, stack_id, **kwargs)
 
+        # set resources removal policy for different environments
+        if deploy_env == "prod":
+            REMOVAL_POLICY = core.RemovalPolicy.RETAIN
```
l0b0, line 31: Local variables should be lower case.

datalake/processing_stack.py
```diff
@@ -21,12 +22,32 @@
 class ProcessingStack(core.Stack):
     """Data Lake processing stack definition."""
 
-    # pylint: disable=too-many-locals
+    # pylint: disable=too-many-locals,too-complex
```
l0b0, line 25: Can we simplify this code to avoid this complexity? For example, there are at least three places where we check `if deploy_env == "prod"` - those can all be joined into a single `if`/`else` to set some configuration.

l0b0 (changes requested):

datalake/processing_stack.py
imincik: Should be fixed in 17923d87546f6713a687ded8ccb111eff35e4b34

datalake/processing_stack.py
imincik: Should be fixed in 17923d87546f6713a687ded8ccb111eff35e4b34

imincik:
@l0b0 , sorry I pressed "Update with master" button on this page which did merge instead of rebase.

l0b0 (approved):

l0b0:
> @l0b0 , sorry I pressed "Update with master" button on this page which did merge instead of rebase.

Yeah, I've used that by accident as well. Unfortunately there doesn't seem to be a setting to do a rebase rather than merging in master, making for a pretty terrible version history.

PR #186, opened 2021-01-20
ci: Use recommended APT install commands in CI

l0b0:


.github/workflows/test.yml
```diff
@@ -65,7 +65,11 @@ jobs:
                   submodules: 'true'
 
             - name: Install AWS CLI
-              run: sudo apt install awscli
+              run: |
+                  sudo apt-get update
```
imincik, line 69: I would move `apt-get update` to separate task. Thanks

imincik (changes requested):

PR #188, opened 2021-01-20
refactor(infra): use constants for public stack resources

imincik:
This change is based on #187


l0b0 (changes requested):
Does not yet pass CI.

imincik:
No relevant anymore.

PR #198, opened 2021-01-27
ci: Allow deployment of multiple stacks

MitchellPaff:
Modified existing stacks and resources to all be prefixed with the DEPLOY_ENV.

Updated README to reflect that a developer should sit this to their username in order to deploy a personal development stack.

Utilised AWS Parameter Store to store the resource names for each stack.


README.md
```diff
@@ -96,12 +96,17 @@ Re-run `. .venv/bin/activate` in each shell.
...
-    string without spaces. Default: dev.
-* **DATALAKE_USE_EXISTING_VPC:** determine if networking stack will use existing VPC in target AWS
-    account or will create a new one. Existing VPC must to be used must contain following tags -
-    "ApplicationName": "geospatial-data-lake", "ApplicationLayer": "networking".
-    Allowed values: true, false. Default: false.
+   * **`DEPLOY_ENV`:** set deployment environment. 
+     For your personal development stack: set DEPLOY_ENV to your username. 
+     
+     ```bash
+     export DEPLOY_ENV=yourusername
```
l0b0, line 107: You could suggest `export DEPLOY_ENV="$USER"` to simplify this a bit.

datalake/api_stack.py
```diff
@@ -8,7 +8,9 @@
 class APIStack(core.Stack):
     """Data Lake stack definition."""
 
-    def __init__(self, scope: core.Construct, stack_id: str, datasets_table, **kwargs) -> None:
+    def __init__(
+        self, scope: core.Construct, stack_id: str, deploy_env, datasets_table, **kwargs
```
l0b0, line 12: `deploy_env` should be type `str`.

app.py
```diff
@@ -14,7 +14,7 @@
 from datalake.staging_stack import StagingStack
 from datalake.storage_stack import StorageStack
 
-ENV = os.environ.get("DEPLOY_ENV", "dev")
+ENV = os.environ["DEPLOY_ENV"]
```
l0b0, line 17: We could now import this from `parameter_manager`.

datalake/backend/parameter_manager.py
```diff
@@ -0,0 +1,21 @@
...
+
+import boto3
+
+ssm_client = boto3.client("ssm")
+
+ENV = os.environ["DEPLOY_ENV"]
+
+
+class ResourceName(Enum):
+    DatasetsTableName = "DatasetsTableName"
```
l0b0, line 12: If we use `DatasetsTableName = auto()` we can use for example `ResourceName.DatasetsTableName.name` to get the string name of that entry.

Also, the entry names should probably be `UPPER_CASE` since they are effectively constants.

datalake/processing_stack.py
```diff
@@ -33,6 +35,14 @@ def __init__(
             self, "processing_assets", deploy_env=deploy_env, application_layer=application_layer
         )
 
+        aws_ssm.StringParameter(
+            self,
+            ResourceName.ProcessingAssetsTableName.value,
+            parameter_name=f"/{deploy_env}/{ResourceName.ProcessingAssetsTableName.value}",
+            description="Name of the Processing Assets Table",
```
l0b0, line 42: Considering the parameter name, does this description add anything? Ditto for the other parameters.

l0b0 (changes requested):

app.py
MitchellPaff: No longer applicable.

datalake/backend/endpoints/parameter_manager.py
```diff
@@ -0,0 +1,21 @@
...
+    DATASETS_TABLE_NAME = auto()
+    PROCESSING_ASSETS_TABLE_NAME = auto()
+    STORAGE_BUCKET_NAME = auto()
+    DATASET_STAGING_BUCKET_NAME = auto()
+
+
+def get_param(parameter: ResourceName):
+    path = f"/{ENV}/{parameter.value}"
+    ssm_parameter = ssm_client.get_parameter(Name=path)
+    return ssm_parameter["Parameter"]["Value"]
```
l0b0, line 21: We might want to do something like `assert ssm_parameter["Result"] == success, ssm_parameter` to get some details if the request failed.

datalake/backend/endpoints/parameter_manager.py
```diff
@@ -16,6 +17,9 @@ class ResourceName(Enum):
...
 def get_param(parameter: ResourceName):
-    path = f"/{ENV}/{parameter.value}"
-    ssm_parameter = ssm_client.get_parameter(Name=path)
-    return ssm_parameter["Parameter"]["Value"]
+    path = f"/{ENV}/{parameter.name}"
+
+    try:
+        return ssm_client.get_parameter(Name=path)["Parameter"]["Value"]
+    except ClientError as e:
+        print(e.response["Error"]["Code"])
```
l0b0, line 25: I don't think we want to ignore this exception. If any kind of error (permission, no such parameter, etc) is going to result in a `ClientError` then we should simply not catch it and let the application crash, because it would be an implementation error. [My original suggestion](https://github.com/linz/geospatial-data-lake/pull/198#discussion_r565647975) was made under the assumption that `get_parameter` might not throw an exception in certain error cases such as the parameter not existing. If that is not the case we can simply remove the error handling.

datalake/backend/endpoints/utils.py
```diff
@@ -13,3 +14,10 @@ def success_response(code, body):
     """Return success response content as string."""
 
     return {"statusCode": code, "body": body}
+
+
+class ResourceName(Enum):
+    DATASETS_TABLE_NAME = "datasets"
```
l0b0, line 20: Could we add `DEPLOY_ENV` to the start of these names *here?* We'll need to move the constant from app.py here to avoid a circular import.

datalake/backend/tests/test_lambda_functions.py
```diff
@@ -3,10 +3,13 @@
 """
 
 import json
+import os
 import uuid
 
 from pytest import mark
 
+ENV = os.environ["DEPLOY_ENV"]
```
l0b0, line 11: We should be able to pull this environment variable out in only one place, and import it elsewhere.

l0b0 (changes requested):

l0b0 (approved):
Excellent, thanks!

l0b0 (changes requested):
As discussed, datalake/networking_stack.py looks up the VPC by tags which are no longer unique.

README.md
```diff
@@ -100,17 +100,22 @@ Re-run `. .venv/bin/activate` in each shell.
...
+     export DEPLOY_ENV="$USER"
+     ```
+     Other values ued by CI pipelines include: prod, nonprod, ci, dev or any string without spaces. Default: dev.
 1. Bootstrap CDK (only once per profile)
 
     ```bash
     cdk --profile=<AWS-PROFILE-NAME> bootstrap aws://unknown-account/ap-southeast-2
     ```
+1. VPC pre-requisite
+   - A VPC must be created manually in your AWS account with the following tags:
```
l0b0, line 116: This should probably not be a list with a single entry.

datalake/backend/endpoints/utils.py
```diff
@@ -13,3 +17,13 @@ def success_response(code, body):
     """Return success response content as string."""
 
     return {"statusCode": code, "body": body}
+
+
+class ResourceName(Enum):
+    DATASETS_TABLE_NAME = f"{ENV}-datasets"
+    DATASETS_TITLE_INDEX_NAME = "datasets_title"
+    DATASETS_OWNING_GROUP_INDEX_NAME = "datasets_owning_group"
```
l0b0, line 25: As discussed, these two should probably be separate constants.

datalake/constructs/batch_job_queue.py
```diff
@@ -67,15 +66,27 @@ def __init__(
...
             launch_template_data=launch_template_data,
         )
         launch_template = aws_batch.LaunchTemplateSpecification(
             launch_template_name=cloudformation_launch_template.launch_template_name
         )
 
+        # use existing VPC in LINZ AWS account.
+        # VPC with these tags is required to exist in AWS account before being deployed.
+        # A VPC will not be deployed by this project.
+        datalake_vpc = aws_ec2.Vpc.from_lookup(
```
l0b0, line 79: The "datalake" prefix is probably redundant, since the code is inside the context of the datalake.

datalake/backend/endpoints/utils.py
MitchellPaff: resolved this.

imincik (changes requested):
@MitchellPaff ,

I haven't done in-line comments because personally, I don't agree with some fundamental principals of this change, so it doesn't make a sense for me to go to details:

1. if we want to drop networking stack from our product completely, we should document all mandatory networking details of VPC which are: required subnet types, their sizes, internet connectivity configuration, ... otherwise person deploying this product might experience hard times during deployment. That's why I like much more to keep networking stack which is configured in the way that we know that it should work just fine. In case some different networking configuration is needed by some third party, deployment of networking stack can by omitted. 

2. I would prefer not setting resource names when they are not needed and leave to be auto-generated by AWS. This can save us from troubles during stack updates (see for example: https://aws.amazon.com/premiumsupport/knowledge-center/cloudformation-custom-name/)

3. I would suggest that this change should solve these kind of problems as well - https://github.com/linz/geospatial-data-lake/blob/master/datalake/backend/tests/test_processing_stack.py#L51

Thank you

README.md
```diff
@@ -100,12 +100,13 @@ Re-run `. .venv/bin/activate` in each shell.
...
-    account or will create a new one. Existing VPC must to be used must contain following tags -
-    "ApplicationName": "geospatial-data-lake", "ApplicationLayer": "networking".
-    Allowed values: true, false. Default: false.
+   * **`DEPLOY_ENV`:** set deployment environment. 
+     For your personal development stack: set DEPLOY_ENV to your username. 
+     
+     ```bash
+     export DEPLOY_ENV="$USER"
+     ```
+     Other values ued by CI pipelines include: prod, nonprod, ci, dev or any string without spaces. Default: dev.
```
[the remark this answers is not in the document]
MitchellPaff: Hey @imincik , thanks for your comments.
1.  I agree that this should be documented.  My intent with the change was to remove different code paths for LINZ/non-LINZ developers that are controlled by environment variables. I think this adds unnecessary complexity. I've added the networking_stack.py back in to the project to document the networking requirements and added steps into the readme to help potential other users of the application.  [commit of changes](https://github.com/linz/geospatial-data-lake/pull/198/commits/20d519c9045ca491e97d9d2b36ca9c0bd52b26f8)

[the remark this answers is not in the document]
l0b0: Are we ever running this networking stack code? As in, all of it? If not I would strongly recommend removing it, since the odds that it is actually working and completely in sync with our networking stack is practically nil.

datalake/backend/tests/test_check_stac_metadata.py
```diff
@@ -47,7 +47,7 @@
     },
 }
 
-BUCKET_NAME = f"linz-geospatial-data-lake-{ENV}"
+BUCKET_NAME = ResourceName.STORAGE_BUCKET_NAME.value
```
l0b0, line 50: I would just inline (remove) the constant since it's just a proxy for another constant now.

l0b0 (approved):
Thank you!

datalake/constructs/batch_job_queue.py
```diff
@@ -2,6 +2,9 @@
 
 from aws_cdk import aws_batch, aws_dynamodb, aws_ec2, aws_iam, core
 
+APPLICATION_NAME_TAG_NAME = "ApplicationName"
```
imincik, line 5: I don't think this file is a good place for setting `APPLICATION_NAME_TAG_NAME` and `APPLICATION_NAME` variables. I would use for example `..endpoints.utils` instead.

imincik (approved):
@MitchellPaff , I approve this PR, but I left some comments. Would you please address them in new PR ? 

Also, as we discussed, it would be good to do also:

* use AWS generated resource names instead of custom ones
* remove resource detection function like the one which is detecting state machine name in tests

Thank you

PR #202, opened 2021-01-28
refactor: Reference app.py constants

l0b0:
Pull CDK into a `main` method to allow side-effect-free import of the
constants.


imincik (approved):
Thanks @l0b0

imincik (approved):

PR #267, opened 2021-02-16
feat: Set log level to debug across state machine

l0b0:


imincik (approved):
Looks good to me Victor. Good idea.

Issue #254, opened 2021-02-14
Bug: validate all files as collections


l0b0:
### Bug Description

We only want to support collection STAC files for now, so we should remove the catalog validation code.

billgeo:
As discussed, I think we should enforce the best practice guide as described https://github.com/radiantearth/stac-spec/blob/master/best-practices.md and enforce STAC files to be named either `catalog.json`, `collection.json` or `<item_name>.json`.

billgeo:
And also happy that we only accept collection.json files for the MVP.

l0b0:
As far as I can tell we are already doing this. See https://github.com/linz/geospatial-data-lake/pull/284 for why we can't remove all reference to the catalog schema.

PR #284, opened 2021-02-21
refactor: Remove mentions of catalog STAC schema

l0b0:
We are going to treat all files as STAC collections for now.

Closes https://github.com/linz/geospatial-data-lake/issues/254.


billgeo:
LGTM from a business point of view

l0b0:
Closing since without this schema in the schema store jsonschema ends up downloading the catalog schema from the Internet rather than using the local file. Possibly the `$.allOf.$ref` is taken to be relative to `$.$id`?

PR #308, opened 2021-02-25
feat: Import dataset task

MitchellPaff:
Task in step function to copy dataset from staging bucket to storage after validation success.

minor rework of other classes to facilitate the storage of metadata files in the processing-assets table


datalake/api_stack.py
```diff
@@ -59,7 +60,7 @@ def __init__(  # pylint: disable=too-many-arguments
             if endpoint == "dataset_versions":
                 state_machine_parameter = aws_ssm.StringParameter.from_string_parameter_attributes(
                     self,
-                    "StepFunctionStateMachineARN",
+                    "step-func-statemachine-arn",
```
l0b0, line 63: If this *needs* to be the same as the `aws_ssm.StringParameter(` `id` parameter we should make it a constant. Otherwise we should make them different, to avoid confusion. In any case they should be different from the constant value.

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+JsonObject = MutableMapping[str, Any]
+
+ssm_client = boto3.client("ssm")
+sts_client = boto3.client("sts")
+s3_client = boto3.client("s3")
+s3control_client = boto3.client("s3control")
+
+ENV = environ.get("DEPLOY_ENV", "test")
+STORAGE_BUCKET_PARAMETER = f"/{ENV}/storage-bucket-arn"
+S3_BATCH_COPY_ROLE_PARAMETER = f"/{ENV}/s3-batch-copy-role-arn"
```
l0b0, line 26: These should probably be have a `_NAME` suffix to distinguish the name from the value.

datalake/backend/tests/utils.py
```diff
@@ -239,6 +244,45 @@ def __exit__(
         self._item.delete()
 
 
+class ProcessingAsset:
+    def __init__(
+        self,
+        asset_id: Optional[str] = None,
+        item_index: Optional[str] = None,
+        multihash: Optional[str] = None,
+        url: Optional[str] = None,
```
l0b0, line 253: All of the calls to this method include a `url` parameter, so it shouldn't be optional.

datalake/backend/tests/utils.py
```diff
@@ -239,6 +244,45 @@ def __exit__(
         self._item.delete()
 
 
+class ProcessingAsset:
+    def __init__(
+        self,
+        asset_id: Optional[str] = None,
```
l0b0, line 250: All of the calls to this method include an `asset_id` parameter, so it shouldn't be optional.

datalake/backend/tests/utils.py
```diff
@@ -239,6 +244,45 @@ def __exit__(
...
+        item_index: Optional[str] = None,
```
l0b0, line 251: None of the calls to this method specify `item_index`, so this argument should probably be removed.

datalake/constructs/bundled_lambda_function.py
```diff
@@ -33,6 +34,7 @@ def __init__(
             handler=f"processing.{directory}.task.lambda_handler",
             runtime=aws_lambda.Runtime.PYTHON_3_8,
             environment=environment,
+            timeout=Duration.seconds(60),
```
l0b0, line 37: We should probably add a comment to explain why not the default 3s and why not the max 15 minutes. Since this pattern is used in more than one place, it might make sense to document this in CODING.md?

datalake/processing_stack.py
```diff
@@ -157,6 +163,95 @@ def __init__(
             application_layer=application_layer,
         ).lambda_invoke
 
+        storage_bucket_arn = aws_ssm.StringParameter.from_string_parameter_attributes(
```
l0b0, line 166: I would add `_name` to the variable name to avoid this being confused for the actual ARN value variable.

datalake/processing_stack.py
```diff
@@ -157,6 +163,95 @@ def __init__(
...
+        )
+        s3_batch_copy_role.add_to_policy(
+            aws_iam.PolicyStatement(
+                actions=[
+                    "s3:GetObject",
+                    "s3:GetObjectAcl",
+                    "s3:GetObjectTagging",
+                ],
+                resources=[
+                    staging_bucket_arn.string_value + "/*",
```
l0b0, line 197: This should be a format string.

datalake/processing_stack.py
```diff
@@ -157,6 +163,95 @@ def __init__(
...
+                actions=[
+                    "s3:PutObject",
+                    "s3:PutObjectAcl",
+                    "s3:PutObjectTagging",
+                    "s3:GetObject",
+                    "s3:GetObjectVersion",
+                    "s3:GetBucketLocation",
+                ],
+                resources=[
+                    storage_bucket_arn.string_value + "/*",
```
l0b0, line 212: This should be a format string.

datalake/staging_stack.py
```diff
@@ -23,3 +27,11 @@ def __init__(self, scope: core.Construct, stack_id: str, **kwargs: Any) -> None:
...
         )
         Tags.of(self.staging_bucket).add("ApplicationLayer", "storage")  # type: ignore[arg-type]
+
+        aws_ssm.StringParameter(
+            self,
+            "staging-bucket-arn",
+            description=f"Staging Bucket ARN for {deploy_env}",
+            parameter_name=STAGING_BUCKET_PARAMETER,
+            string_value=self.staging_bucket.bucket_arn,
+        )
```
l0b0, line 37: As discussed, it looks like this is only used elsewhere in CDK, not at runtime. We should avoid this overhead and complexity.

app.py
```diff
@@ -39,13 +39,23 @@ def main() -> None:
...
+
     processing = ProcessingStack(
         app,
         "processing",
         stack_name=f"{ENV}-geospatial-data-lake-processing",
         env={"region": region, "account": account},
         deploy_env=ENV,
     )
+    processing.add_dependency(storage)
+    processing.add_dependency(staging)
```
l0b0, line 58: Why do we have to wait for these? Is it possible to rearrange things so that they can continue to be deployed in parallel?

l0b0 (changes requested):

datalake/backend/tests/test_import_dataset_logging.py
```diff
@@ -0,0 +1,126 @@
...
+class TestLogging:
+    logger: logging.Logger
+
+    @classmethod
+    def setup_class(cls) -> None:
+        cls.logger = logging.getLogger("datalake.backend.processing.import_dataset.task")
+
+    @mark.infrastructure
+    @patch("datalake.backend.processing.import_dataset.task.s3control_client.create_job")
+    def test_should_log_payload(self, create_job_mock: MagicMock) -> None:
```
l0b0, line 30: I think if you `patch("datalake.backend.processing.import_dataset.task.validate")` (probably at line 41, not here, since we shouldn't need the validate mock for anything) this doesn't need to be `@mark.infrastructure`, meaning it'll run much faster and can run locally without deploying.

datalake/backend/tests/test_import_dataset_logging.py
```diff
@@ -0,0 +1,126 @@
...
+        # Given
+
+        create_job_mock.return_value = {"JobId": "Some Response"}
```
l0b0, line 33: Do you need to specify the return value?

datalake/backend/tests/test_import_dataset_logging.py
```diff
@@ -0,0 +1,126 @@
...
+
+        body = {}
+        body["dataset_id"] = any_dataset_id()
+        body["metadata_url"] = any_s3_url()
+        body["version_id"] = any_dataset_version_id()
+        expected_payload_log = dumps({"payload": body})
+
+        with patch.object(self.logger, "debug") as log_mock:
```
l0b0, line 41: I'd probably call this the `logger_mock`, since it's not the log itself.

datalake/backend/tests/test_import_dataset_logging.py
```diff
@@ -0,0 +1,126 @@
...
+
+            # When
+            lambda_handler(body, any_lambda_context())
+
+            # Then
+            log_mock.assert_any_call(expected_payload_log)
+
+    @patch("datalake.backend.processing.import_dataset.task.validate")
+    def test_should_log_missing_argument_warning(self, validate_schema_mock: MagicMock) -> None:
```
l0b0, line 50: This title doesn't correspond to the test contents. It looks instead like it logs any validation error.

datalake/backend/tests/test_import_dataset_logging.py
```diff
@@ -0,0 +1,126 @@
...
+                lambda_handler(body, any_lambda_context())
+
+                # Then
+                with subtests.test():
+                    log_mock.assert_any_call(expected_asset_log)
+                with subtests.test():
+                    log_mock.assert_any_call(expected_metadata_log)
+
+    @patch("datalake.backend.processing.import_dataset.task.s3control_client.create_job")
+    def test_should_log_s3_batch_repsonse(self, create_job_mock: MagicMock) -> None:
```
l0b0, line 109: Typo in `repsonse`.

datalake/backend/tests/test_processing_stack.py
```diff
@@ -34,14 +35,23 @@
 
 
 @mark.infrastructure
-def test_should_create_state_machine_arn_parameter(ssm_client: SSMClient) -> None:
+def test_should_retrieve_state_machine_arn_parameter(ssm_client: SSMClient) -> None:
```
l0b0, line 38: As discussed, this test name could probably be better phrased as "state machine ARN parameter should exist". Ditto below.

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+    metadata_url = payload["metadata_url"]
+
+    storage_bucket_arn = get_param(STORAGE_BUCKET_PARAMETER)
+    storage_bucket_name = storage_bucket_arn.rsplit(":", maxsplit=1)[-1]
+
+    staging_bucket_name = urlparse(metadata_url).netloc
+    manifest_key = f"manifests/{dataset_version_id}.csv"
+
+    with smart_open(f"s3://{storage_bucket_name}/{manifest_key}", "w") as s3_manifest:
+        for file in ProcessingAssetsModel.query(
```
l0b0, line 63: `file` is a very generic word. Also, `query` returns an iterable of items, so that might be a better name.

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+            f"DATASET#{dataset_id}#VERSION#{dataset_version_id}"
+        ):
+            logger.debug(json.dumps({"Adding file to manifest": file.url}, default=str))
```
l0b0, line 66: `default=str` should only be necessary when serializing objects which `dumps` would normally refuse to serialize. I don't think it's needed here.

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+            key = urlparse(file.url).path[1:]
+            s3_manifest.write(f"{staging_bucket_name},{key}\n")
+
+    # trigger s3 batch copy operation
+    response = s3control_client.create_job(
+        AccountId=sts_client.get_caller_identity()["Account"],
```
l0b0, line 72: If `get_caller_identity` fails we might only get a `KeyError` message here with no more context. Something like `assert "Account" in caller_identity := sts_client.get_caller_identity(), caller_identity` (not sure if that syntax actually works) might be appropriate.

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+            "Spec": {
+                "Format": "S3BatchOperations_CSV_20180820",
+                "Fields": [
+                    "Bucket",
+                    "Key",
+                ],
+            },
+            "Location": {
+                "ObjectArn": f"{storage_bucket_arn}/{manifest_key}",
+                "ETag": s3_client.head_object(Bucket=storage_bucket_name, Key=manifest_key)["ETag"],
```
l0b0, line 90: As for `get_caller_identity`, it probably makes sense to pull this out to make sure we get more context if the object doesn't exist.

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+            },
+        },
+        Report={
+            "Enabled": True,
+            "Bucket": storage_bucket_arn,
+            "Format": "Report_CSV_20180820",
```
l0b0, line 96: Is this format value a constant we can refer to?

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+                "TargetResource": storage_bucket_arn,
+                "TargetKeyPrefix": f"{dataset_id}/{dataset_version_id}",
+            }
+        },
+        Manifest={
+            "Spec": {
+                "Format": "S3BatchOperations_CSV_20180820",
+                "Fields": [
+                    "Bucket",
+                    "Key",
```
l0b0, line 85: This list can be on a single line. The trailing comma prevents `black` from auto-formatting it that way.

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+            "Bucket": storage_bucket_arn,
+            "Format": "Report_CSV_20180820",
+            "Prefix": f"reports/{datetime.now()}-{dataset_version_id}",
+            "ReportScope": "AllTasks",
+        },
+        Priority=1,
+        RoleArn=get_param(S3_BATCH_COPY_ROLE_PARAMETER),
+        ClientRequestToken=uuid4().hex,
+    )
+    logger.debug(json.dumps({"s3 batch response": response}, default=str))
```
l0b0, line 104: Would it be possible to log the request object as well?

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+
+    return success_response(200, {"job_id": response["JobId"]})
+
+
+def set_up_logging() -> logging.Logger:
```
l0b0, line 109: This is probably a good candidate for a utility function which would just need to take the `__name__` as a parameter.

l0b0 (changes requested):

imincik:
@MitchellPaff , it looks like tests are not working for some reason. They are stuck in `Waiting for status to be reported` state.

datalake/backend/processing/import_dataset/task.py
MitchellPaff: yes

datalake/backend/tests/utils.py
MitchellPaff: - [ ]

datalake/backend/processing/import_dataset/task.py
l0b0: Should we refer to that constant instead, in that case? ;)

datalake/backend/processing/import_dataset/task.py
```diff
@@ -0,0 +1,142 @@
...
+    return {"statusCode": code, "body": {"message": f"{http_responses[code]}: {message}"}}
+
+
+def success_response(code: int, body: Union[JsonList, JsonObject]) -> JsonObject:
+    """Return success response content as string."""
+
+    return {"statusCode": code, "body": body}
+
+
+def get_param(parameter: str) -> str:
```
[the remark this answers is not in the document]
MitchellPaff: I ran into issues removing this into utils class due to it using the boto3 ssm_client. Could give it another go

datalake/backend/processing/import_dataset/task.py
MitchellPaff: Sorry, didn't understand what you meant. It's a magic string and can't see it surface anywhere from boto3 s3_control client

datalake/backend/processing/import_dataset/task.py
MitchellPaff: Don't think so.

datalake/backend/tests/test_import_dataset.py
```diff
@@ -0,0 +1,139 @@
...
+        any_safe_filename(),
+    ) as asset_s3_object:
+        metadata_stac_object = deepcopy(MINIMAL_VALID_STAC_OBJECT)
+        metadata_stac_object["assets"] = {
+            any_stac_asset_name(): {
+                "href": asset_s3_object.url,
+                "checksum:multihash": first_asset_multihash,
+            },
+        }
+        metadata_content = dumps(metadata_stac_object).encode()
```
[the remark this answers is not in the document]
MitchellPaff: I don't understand the extra parameters required for `dump`

datalake/backend/processing/import_dataset/task.py
MitchellPaff: I think it's a bit extraneous to be asserting that the AWS libraries are returning what they say they will.

datalake/backend/tests/test_import_dataset.py
```diff
@@ -0,0 +1,139 @@
...
+                # When
+                body = {}
+                body["dataset_id"] = dataset_id
+                body["version_id"] = version_id
+                body["metadata_url"] = metadata_s3_object.url
+                body["type"] = any_valid_dataset_type()
+
+                response = lambda_handler(body, any_lambda_context())
+
+                final_states = ["Complete", "Failed", "Cancelled"]
```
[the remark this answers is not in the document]
MitchellPaff: No constant values in boto3  available to us for this. Do you mean creating an Enum?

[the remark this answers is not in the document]
l0b0: No, I was hoping that the Python API code would expose constants. Otherwise we're hardcoding values which belong to *their* API, which should be avoided whenever possible since those values could change.

datalake/backend/processing/import_dataset/task.py
l0b0: That argument is only valid if the call is *guaranteed* to succeed, which can only be the case if the call is local. This call can fail for any number of reasons, like a broken network connection, missing credentials, a change in the API, or an API outage. If we just do `AccountId=sts_client.get_caller_identity()["Account"]` we'll have no way of figuring out what went wrong if that call ever fails.

datalake/backend/tests/test_import_dataset.py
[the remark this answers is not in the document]
l0b0: Oh, right, it writes to a file pointer. Sorry, my bad.

datalake/backend/processing/import_dataset/task.py
l0b0: Oh, I meant whether there was a constant in boto3 or other AWS library with this value, so we could refer to that rather than relying on yet another magic string. I had a quick look and agree, so we can just leave it.

datalake/backend/processing/import_dataset/task.py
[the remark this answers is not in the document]
l0b0: You could pass the client as an argument to the function, does that help?

backend/import_dataset/task.py
```diff
@@ -0,0 +1,110 @@
...
+    with smart_open(f"s3://{storage_bucket_name}/{manifest_key}", "w") as s3_manifest:
+        for item in ProcessingAssetsModel.query(
+            f"DATASET#{dataset_id}#VERSION#{dataset_version_id}"
+        ):
+            logger.debug(json.dumps({"Adding file to manifest": item.url}))
+            key = urlparse(item.url).path[1:]
+            s3_manifest.write(f"{staging_bucket_name},{key}\n")
+
+    caller_identity = STS_CLIENT.get_caller_identity()
+    assert "Account" in caller_identity
```
l0b0, line 68: This should probably have `, caller_identity` at the end so that the error message includes the actual response.

backend/import_dataset/task.py
```diff
@@ -0,0 +1,110 @@
...
+    account_number = caller_identity["Account"]
+
+    manifest_s3_object = S3_CLIENT.head_object(Bucket=storage_bucket_name, Key=manifest_key)
+    assert "ETag" in manifest_s3_object
```
l0b0, line 72: This should probably have `, manifest_s3_object` at the end so that the error message includes the actual response.

l0b0 (changes requested):

l0b0 (approved):

PR #320, opened 2021-03-01
docs: usage for staging bucket permissions

billgeo:
I had trouble setting up permissions from my test bucket, so thought it would be good to have some helpful pointers for this. I ended up using this policy, although not sure it is the best/safest example?


USAGE.md
```diff
@@ -1,16 +1,38 @@
...
+        {
+            "Sid": "data-lake-readonly",
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": "arn:aws:iam::<DATALAKE_AWS_ACCOUNT_ID>:root"
+            },
+            "Action": [
+                "s3:GetObject"
+            ],
+            "Resource": "arn:aws:s3:::<YOUR_STAGING_BUCKET>/<YOUR_DATASET>"
```
imincik, line 30: I think it should be `Resource": "arn:aws:s3:::<YOUR_STAGING_BUCKET>/<YOUR_DATASET>/*` (adding slash and asterisk at the end)

billgeo: Thanks. Fixed that.

billgeo:
@imincik, please re-review.

imincik (approved):
Thanks @billgeo

Issue #334, opened 2021-03-03
Use "should_" test function prefix for test discovery


l0b0:
### Enabler

So that test names are shorter but still descriptive, we want to change the Pytest configuration to use `should_` as the test function prefix.

#### Acceptance Criteria

- [ ] Test function names start with `should_`
- [ ] Pytest still runs all the same tests as before.

#### Additional context

We already use `test_should_` as the de facto prefix.

#### Tasks

- [ ] [Configure Pytest](https://docs.pytest.org/en/stable/example/pythoncollection.html#changing-naming-conventions)
- [ ] Rename test functions.

#### Definition of Ready
- [ ] This story is __ready__ to work on, according to the [team's definition](https://confluence.linz.govt.nz/pages/viewpage.action?pageId=87930423)

#### Definition of Done
- [ ] This story is __done__, according to the [team's definition](https://confluence.linz.govt.nz/pages/viewpage.action?pageId=87930423)

PR #362, opened 2021-03-09
refactor: Use "should_" test name prefix

l0b0:
Closes https://github.com/linz/geospatial-data-lake/issues/334


imincik (approved):
Should be OK. Thanks

PR #367, opened 2021-03-09
feat: Get Import Status

MitchellPaff:
Lambda endpoint to allow user to retrieve status update for Step Function and S3 batch Copy operation for their dataset.

### DIFF
--- a/tests/test_import_status_logging.py
+++ b/tests/test_import_status_logging.py
@@ -7,7 +7,7 @@ from jsonschema import ValidationError  # type: ignore[import]
 
 from backend.import_status.get import get_import_status, get_s3_batch_copy_status
 
-from .general_generators import any_valid_arn
+from .aws_utils import any_arn_formatted_string
 
 
 class TestLogging:
@@ -23,7 +23,7 @@ class TestLogging:
         # Given
         payload = {
             "httpMethod": "GET",
-            "body": {"execution_arn": any_valid_arn()},
+            "body": {"execution_arn": any_arn_formatted_string()},
         }
 
         expected_payload_log = dumps({"payload": payload})
@@ -59,7 +59,7 @@ class TestLogging:
             # Then
             logger_mock.assert_any_call(expected_log)
 
-    @patch("backend.import_status.get.STEPFUNCTIONS_CLIENT.describe_execution")
+    @patch("backend.import_status.get.STEP_FUNCTIONS_CLIENT.describe_execution")
     def test_should_log_stepfunctions_status_response(
         self,
         describe_execution_mock: MagicMock,
@@ -77,7 +77,7 @@ class TestLogging:
             get_import_status(
                 {
                     "httpMethod": "GET",
-                    "body": {"execution_arn": any_valid_arn()},
+                    "body": {"execution_arn": any_arn_formatted_string()},
                 }
             )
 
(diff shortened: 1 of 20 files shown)


tests/test_import_status_logging.py
```diff
@@ -0,0 +1,106 @@
...
+        describe_s3_job_mock.return_value = s3_batch_response = {"Job": {"Status": "Some Response"}}
+        expected_response_log = json.dumps({"s3 batch response": s3_batch_response})
+
+        with patch.object(self.logger, "debug") as logger_mock, patch(
+            "backend.import_status.get.STS_CLIENT.get_caller_identity"
+        ):
+            get_s3_batch_copy_status(
+                "test",
+                self.logger,
+            )
```
l0b0, line 103: This can just be one line rather than four. Black doesn't concatenate lines when the last one ends with a comma, presumably to avoid destroying tuples like

```python
(
    some_very_long_value,
)
```

tests/test_processing_stack.py
```diff
@@ -165,3 +167,16 @@ def should_successfully_run_dataset_version_creation_process(
...
+                    FunctionName=ResourceName.IMPORT_STATUS_ENDPOINT_FUNCTION_NAME.value,
+                    Payload=json.dumps(
+                        {"httpMethod": "GET", "body": {"execution_arn": execution["executionArn"]}}
+                    ).encode(),
+                    InvocationType="RequestResponse",
+                )
+                status_json_resp = json.load(status_resp["Payload"])
+                assert status_json_resp["statusCode"] == 200, status_json_resp
+                assert status_json_resp["body"]["validation_status"] == "SUCCEEDED"
+                assert status_json_resp["body"]["upload_status"] == "Complete"
```
l0b0: Each of these assertions should be in separate subtests, *or* the test should do a single assert on the whole response.

tests/test_processing_stack.py
```diff
@@ -165,3 +167,21 @@ def should_successfully_run_dataset_version_creation_process(
                 time.sleep(5)
 
             assert copy_job["Job"]["Status"] == "Complete", copy_job
+
+            with subtests.test(msg="Import Status Endpoint"):
```
l0b0, line 171: This test is getting so complicated that we should probably split it into three:

- Should finish the state machine successfully
- Should complete S3 batch copy
- Should report import status after success

l0b0 (changes requested):

tests/test_import_status_logging.py
```diff
@@ -0,0 +1,106 @@
...
+            "httpMethod": "GET",
+            "body": {"execution_arn": any_valid_arn()},
+        }
+
+        expected_payload_log = dumps({"payload": payload})
+
+        with patch.object(self.logger, "debug") as logger_mock, patch(
+            "backend.import_status.get.validate"
+        ) as validate_mock:
+            validate_mock.side_effect = ValidationError("test")
```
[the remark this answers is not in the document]
MitchellPaff: No - trying to avoid having to mock extensively in test by ending the test short.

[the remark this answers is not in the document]
l0b0: It might be simpler to use something like `validate_mock.return_value = None` in that case?

l0b0 (approved):

PR #371, opened 2021-03-10
fix: Run non-infrastructure tests offline

l0b0:
Based on https://github.com/linz/geospatial-data-lake/pull/370


MitchellPaff (dismissed):

imincik (approved):
@l0b0 , I am OK with this change. My only comment is if moving of `JsonObject` is relevant for this change. If not, then it is probably better to separate it to another PR.

l0b0:
> @l0b0 , I am OK with this change. My only comment is if moving of `JsonObject` is relevant for this change. If not, then it is probably better to separate it to another PR.

Done in https://github.com/linz/geospatial-data-lake/pull/377, cheers!

MitchellPaff (approved):

PR #376, opened 2021-03-10
fix: Use default GitHub merge commit message

l0b0:
The commit message should include all messages from all commits in the
pull request. See <https://github.com/chdsbd/kodiak/issues/622>.


imincik (approved):
Thanks

PR #390, opened 2021-03-11
feat(infra): fixed users role name

imincik:
Fixed Data Lake Users role name is needed to link role with Azure AD group configuration.


backend/resources.py
```diff
@@ -12,3 +12,4 @@ class ResourceName(Enum):
     PROCESSING_ASSETS_TABLE_NAME = f"{ENV}-processing-assets"
     STORAGE_BUCKET_NAME = f"{ENV}-linz-geospatial-data-lake"
     VALIDATION_RESULTS_TABLE_NAME = f"{ENV}-validation-results"
+    USERS_ROLE_NAME = f"{ENV}-data-lake-users"
```
l0b0, line 15: This should be above the previous line to be in order

l0b0 (changes requested):

backend/resources.py
imincik: Thanks, fixed and rebased.

l0b0 (approved):

PR #396, opened 2021-03-12
feat: add extra fields to validation table

MitchellPaff:


backend/validation_results_model.py
```diff
@@ -1,14 +1,40 @@
...
+
 class ValidationResultsModel(Model):
     class Meta:  # pylint:disable=too-few-public-methods
         table_name = ResourceName.VALIDATION_RESULTS_TABLE_NAME.value
         region = "ap-southeast-2"  # TODO: don't hardcode region
 
     pk = UnicodeAttribute(hash_key=True)
     sk = UnicodeAttribute(range_key=True)
-    success = BooleanAttribute()
+    status = UnicodeAttribute()
```
l0b0, line 37: We might want to rename either this or `ValidationResult` to use the same noun.

backend/validation_results_model.py
```diff
@@ -1,14 +1,40 @@
...
+    info = UnicodeAttribute(null=True)
```
l0b0, line 38: Shouldn't this be a dictionary so we can include arbitrary information about errors? If so, `info` seems a bit overly generic, like `data`. Would `details` or some synonym work?  If not, what about calling it `message` instead?

backend/validation_results_model.py
```diff
@@ -1,14 +1,40 @@
...
+):  # pylint:disable=too-few-public-methods,inherit-non-class
+    class Meta:  # pylint:disable=too-few-public-methods
+
+        index_name = "validation_outcome"
+        read_capacity_units = 1
+        write_capacity_units = 1
+        projection = AllProjection()
+
+    pk = UnicodeAttribute(hash_key=True)
+    status = UnicodeAttribute(range_key=True)
```
l0b0, line 27: These are basically copies of the actual index defined in infrastructure/processing_stack.py. As discussed we should probably tie them together if possible, so that we're not reimplementing this.

l0b0 (changes requested):

l0b0 (dismissed):

l0b0 (approved):

PR #405, opened 2021-03-15
refactor: Rename log key

l0b0:
Use the same name as in AWS Lambda: "event".


imincik (approved):
I'm OK with that, thanks.

PR #406, opened 2021-03-15
feat(infra): automatically generated AWS resource names

imincik:


backend/parameter_store.py
```diff
@@ -1,20 +1,46 @@
...
+#     # When type checking we want to use the third party package's stub
+#     from mypy_boto3_ssm import SSMClient
+# else:
+#     # In production we want to avoid depending on a package which has no runtime impact
+#     SSMClient = object
 
-def get_param(parameter: str, ssm_client: SSMClient, logger: logging.Logger) -> str:
+
+# def get_param(parameter: str, ssm_client: SSMClient, logger: logging.Logger) -> str:
+def get_param(parameter: str) -> str:
```
MitchellPaff, line 26: get_param could take as input (parameter: ParameterName) and then we could pass in ParameterName.example instead of ParameterName.example.value everywhere we use it.

PR #409, opened 2021-03-15
feat: get import validation status

MitchellPaff:
Added support for returning validation results to the end user

### DIFF
--- a/backend/import_status/get.py
+++ b/backend/import_status/get.py
@@ -80,9 +80,9 @@ def get_step_function_validation_results(step_function_input: str) -> List[str]:
 
     errors = []
     for error in ValidationResultsModel.validation_outcome_index.query(
-        hash_key, ValidationResultsModel.status == "Failed"
+        hash_key=hash_key, range_key_condition=ValidationResultsModel.result == "Failed"
     ):
-        errors.append(error.info)
+        errors.append(json.dumps(error.details))
 
     return errors
 
--- a/backend/import_status/get.py
+++ b/backend/import_status/get.py
@@ -1,15 +1,14 @@
 """Import Status handler function."""
 import json
 import logging
-from typing import List
 
 import boto3
 from jsonschema import ValidationError, validate  # type: ignore[import]
 
 from ..api_responses import error_response, success_response
 from ..log import set_up_logging
-from ..types import JsonObject
-from ..validation_results_model import ValidationResultsModel
+from ..types import JsonList, JsonObject
+from ..validation_results_model import ValidationResult, ValidationResultsModel
 
 STEP_FUNCTIONS_CLIENT = boto3.client("stepfunctions")
 S3CONTROL_CLIENT = boto3.client("s3control")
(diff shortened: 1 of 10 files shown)


backend/import_status/get.py
```diff
@@ -57,13 +58,40 @@ def get_import_status(event: JsonObject) -> JsonObject:
...
+def get_step_function_validation_results(step_function_input: str) -> JsonList:
+    input_json = json.loads(step_function_input)
+    hash_key = f"DATASET#{input_json['dataset_id']}#VERSION#{input_json['version_id']}"
+
+    errors = []
+    for validation_item in ValidationResultsModel.validation_outcome_index.query(
+        hash_key=hash_key,
+        range_key_condition=ValidationResultsModel.result == ValidationResult.FAILED.value,
+    ):
+        sk_list = validation_item.sk.split("#")
```
l0b0, line 80: You could simplify this to `_, check_type, _, url = validation_item.sk.split("#", maxsplit=4)`. Then the assertion below is unnecessary - if there are fewer than four pieces the assignment will fail.

backend/import_status/get.py
```diff
@@ -57,13 +58,40 @@ def get_import_status(event: JsonObject) -> JsonObject:
...
+        assert len(sk_list) == 4, sk_list
+
+        errors.append(
+            {
+                "check_type": sk_list[1],
```
l0b0, line 85: Could we call this just "check" (same as elsewhere) or "check_name" (maybe more user friendly?)?

backend/import_status/get.py
```diff
@@ -57,13 +58,40 @@ def get_import_status(event: JsonObject) -> JsonObject:
...
+                "result": validation_item.result,
+                "file_url": sk_list[3],
```
l0b0, line 87: Could we call this just "url", same as elsewhere?

backend/import_status/get.py
```diff
@@ -57,13 +58,40 @@ def get_import_status(event: JsonObject) -> JsonObject:
...
+        },
         "upload": upload_response,
     }
 
     return success_response(200, response_body)
 
 
+def get_step_function_validation_results(step_function_input: str) -> JsonList:
+    input_json = json.loads(step_function_input)
+    hash_key = f"DATASET#{input_json['dataset_id']}#VERSION#{input_json['version_id']}"
```
l0b0, line 73: Generally it's considered bad practice to pass an unparsed object to a function and letting it parse it, because it makes the implementation of the inner function rely on the structure of an object retrieved by the outer function. I'd rather pass the dataset and version IDs to this function.

tests/aws_utils.py
```diff
@@ -178,6 +180,36 @@ def __exit__(
...
 
 
+class ValidationItem:
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        asset_id: str,
+        result: ValidationResult,
+        details: JsonObject,
+        url: str,
+        check_type: str,
```
l0b0, line 190: Same as elsewhere, we might want to call this `check_name` or just `check`. We don't have check *types* (yet).

l0b0 (changes requested):

l0b0 (approved):

l0b0 (approved):

PR #419, opened 2021-03-17
fix: report correct status of step function when validation fails

MitchellPaff:
The Step Function not failing when a validation failure is encountered means the code needs to be slightly restructured.


backend/import_status/get.py
```diff
@@ -41,27 +47,25 @@ def get_import_status(event: JsonObject) -> JsonObject:
...
-            "s3_batch_copy" in step_functions_output
-            and "job_id" in step_functions_output["s3_batch_copy"]
-        ), step_function_resp
+    step_func_input = json.loads(step_function_resp["input"])
+    step_functions_output = json.loads(step_function_resp.get("output", "{}"))
 
-        upload_response = get_s3_batch_copy_status(
-            step_functions_output["s3_batch_copy"]["job_id"], LOGGER
-        )
+    validation_status = {
```
l0b0, line 53: I'd probably pull this dictionary out as a constant, to make the code a bit easier to read.

backend/import_status/get.py
```diff
@@ -41,27 +47,25 @@ def get_import_status(event: JsonObject) -> JsonObject:
...
-    if step_function_resp["status"] == "SUCCEEDED":
-        assert "output" in step_function_resp, step_function_resp
-        step_functions_output = json.loads(step_function_resp["output"])
-
-        assert (
-            "s3_batch_copy" in step_functions_output
-            and "job_id" in step_functions_output["s3_batch_copy"]
-        ), step_function_resp
+    step_func_input = json.loads(step_function_resp["input"])
+    step_functions_output = json.loads(step_function_resp.get("output", "{}"))
```
l0b0, line 51: We should probably use the same prefix for both variables, either `step_func` or `step_function`.

l0b0 (changes requested):

l0b0 (approved):

l0b0 (approved):

PR #421, opened 2021-03-17
feat: implement validation handling for incorrect staging bucket access

MitchellPaff:

### DIFF
--- a/backend/check.py
+++ b/backend/check.py
@@ -4,3 +4,4 @@ from enum import Enum
 class Check(Enum):
     CHECKSUM = "checksum"
     JSON_SCHEMA = "JSON schema"
+    STAGING_ACCESS = "Staging bucket access"
--- a/backend/check.py
+++ b/backend/check.py
@@ -4,4 +4,4 @@ from enum import Enum
 class Check(Enum):
     CHECKSUM = "checksum"
     JSON_SCHEMA = "JSON schema"
-    STAGING_ACCESS = "Staging bucket access"
+    STAGING_ACCESS = "staging bucket access"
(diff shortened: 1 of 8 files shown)


backend/check.py
```diff
@@ -4,3 +4,4 @@
 class Check(Enum):
     CHECKSUM = "checksum"
     JSON_SCHEMA = "JSON schema"
+    STAGING_ACCESS = "Staging bucket access"
```
l0b0, line 7: Should probably be lower case like the checksum one.

tests/test_check_files_checksums.py
```diff
@@ -175,3 +160,89 @@ def should_log_error_when_validation_fails(
...
+
+    s3_url = any_s3_url()
+    dataset_id = any_dataset_id()
+    version_id = any_dataset_version_id()
+    hash_key = f"DATASET#{dataset_id}#VERSION#{version_id}"
+
+    array_index = "1"
+
+    # When
+    environ[ARRAY_INDEX_VARIABLE_NAME] = array_index
```
l0b0, line 186: This affects other tests, so we should probably use [`patch.dict`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict) to set it.

Could you also have a look at the other tests which set `environ[…]`? There are a few.

tests/test_check_files_checksums_logging.py
```diff
@@ -49,8 +50,7 @@ def should_log_missing_item(self, subtests: SubTests) -> None:
 
         # When/Then
         with patch.object(self.logger, "error") as logger_mock:
-            with subtests.test(msg="Return code"):
-                assert main() == 1
-
+            with subtests.test(msg="Return code"), raises(DoesNotExist):
```
l0b0, line 53: You can join this line with the previous one, running all three context managers in one.

l0b0 (changes requested):

l0b0 (approved):

PR #423, opened 2021-03-17
feat(infra): use SAML identity provider for users auth

imincik:
This change is introducing Data Lake users authorization using SAML identity provider . It is replacing current authorization based on AWS account ID.

SAML auth is successfully tested manually.

## Issues

Closes: https://github.com/linz/geospatial-data-lake-team/issues/39

PR #434, opened 2021-03-22
docs: README file cleanup

imincik:
This is README file cleanup. Keeps the same content just little bit reorganized with added "Running tests" section.

Based on #423 which needs to be merged before this change.


infrastructure/users_stack.py
```diff
@@ -11,21 +11,25 @@ class UsersStack(Stack):
     def __init__(self, scope: Construct, stack_id: str, **kwargs: Any) -> None:
         super().__init__(scope, stack_id, **kwargs)
 
-        account_ids = (
-            aws_iam.AccountPrincipal(account_id=account_id)
-            for account_id in os.environ.get(
-                "DATALAKE_USERS_AWS_ACCOUNTS_IDS", aws_iam.AccountRootPrincipal().account_id
-            ).split(",")
-        )
+        saml_provider_arn = os.environ.get("DATALAKE_SAML_IDENTITY_PROVIDER_ARN")
```
l0b0, line 14: I don't think this change belongs with a readme cleanup :)

l0b0 (changes requested):

infrastructure/users_stack.py
imincik: Yes it comes from #434

l0b0 (approved):

PR #441, opened 2021-03-23
ci: use different values for SAML Identity provider for different envs

imincik:
There are different ARNs for for `prod` and `nonprod` envs, so they need to be taken from different secret vars.


l0b0 (approved):

.github/workflows/deploy.yml
```diff
@@ -83,7 +83,7 @@ jobs:
         poetry run cdk deploy --all --require-approval never
       env:
         DEPLOY_ENV: nonprod
-        DATALAKE_SAML_IDENTITY_PROVIDER_ARN: ${{ secrets.DATALAKE_SAML_IDENTITY_PROVIDER_ARN }}
+        DATALAKE_SAML_IDENTITY_PROVIDER_ARN: ${{ secrets.DATALAKE_SAML_IDENTITY_PROVIDER_ARN_NON_PROD }}
```
l0b0, line 86: We probably should update the documentation around this at the same time.

l0b0 (changes requested):

.github/workflows/deploy.yml
imincik: Documentation contains SAML configuration.

l0b0: But it refers to the original secret name, not these two.

imincik: But it is the only one Data Lake configuration variable. Because we are deploying to different AWS where SAML ID provider ARN is different we have to use different values in our deployment process. But it doesn't have anything to do with Data Lake configuration as such.

l0b0 (approved):

.github/workflows/deploy.yml
l0b0: Gotcha, thanks!

PR #448, opened 2021-03-24
feat: Add user validation for non S3 URL and multiple directories

MitchellPaff:
## Reference

### DIFF
--- a/tests/test_check_stac_metadata.py
+++ b/tests/test_check_stac_metadata.py
@@ -59,6 +59,109 @@ def should_return_non_zero_exit_code_on_validation_failure(
     assert main() == 1
 
 
+@patch("backend.check_stac_metadata.task.ValidationResultFactory")
+def should_save_non_s3_url_validation_results(
+    validation_results_factory_mock: MagicMock,
+    subtests: SubTests,
+) -> None:
+
+    non_s3_url = any_https_url()
+    dataset_id = any_dataset_id()
+    version_id = any_dataset_version_id()
+
+    sys.argv = [
+        any_program_name(),
+        f"--metadata-url={non_s3_url}",
+        f"--dataset-id={dataset_id}",
+        f"--version-id={version_id}",
+    ]
+
+    with subtests.test(msg="Exit code"):
+        assert main() == 1
+
+    hash_key = f"DATASET#{dataset_id}#VERSION#{version_id}"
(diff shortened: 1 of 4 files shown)


tests/test_check_stac_metadata.py
```diff
@@ -59,6 +59,109 @@ def should_return_non_zero_exit_code_on_validation_failure(
...
+                    ValidationResult.FAILED,
+                    details={"message": f"URL doesn't start with “s3://”: “{non_s3_url}”"},
+                ),
+            ]
+        )
+
+
+@mark.infrastructure
+@patch("backend.check_stac_metadata.task.ValidationResultFactory")
+def should_save_non_multiple_directories_validation_results(
```
l0b0, line 99: Is the "non" part of this name intentional?

tests/test_check_stac_metadata.py
```diff
@@ -59,6 +59,109 @@ def should_return_non_zero_exit_code_on_validation_failure(
...
+    validation_results_factory_mock: MagicMock,
+    subtests: SubTests,
+) -> None:
+    staging_bucket_name = get_param(ParameterName.STAGING_BUCKET_NAME)
+
+    dataset_id = any_dataset_id()
+    version_id = any_dataset_version_id()
+    base_url = f"s3://{staging_bucket_name}/"
+    invalid_url = f"test/{any_safe_filename()}"
```
l0b0, line 108: This should probably be called something like `invalid_key`, since it's not a full URL.

tests/test_check_stac_metadata.py
```diff
@@ -59,6 +59,109 @@ def should_return_non_zero_exit_code_on_validation_failure(
...
+        file_object=json_dict_to_file_object(
+            {
+                **deepcopy(MINIMAL_VALID_STAC_OBJECT),
+                "assets": {
+                    any_asset_name(): {
+                        "href": f"{base_url}{invalid_url}",
+                    }
+                },
+                "links": [
+                    {"href": f"{base_url}{invalid_url}", "rel": "child"},
```
l0b0, line 120: The same URL is used for an asset and a metadata file. Is that intentional? IMO it confuses the assertion.

l0b0 (changes requested):

l0b0 (approved):

PR #450, opened 2021-03-24
docs: USAGE file cleanup

imincik:
Updated Authentication and Authorization instructions and little reordering of content

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -1,25 +1,25 @@
 # Using the Geospatial Data Lake
 
-The purpose of GDL is to store geospatial datasets. This document should provide the technical know-how to create and maintain such datasets.
+The purpose of GDL is to store geospatial datasets. This document should provide the technical
+know-how to create and maintain such datasets.
 
-The keywords "must", "must not", "required", "shall", "shall not", "should", "should not", "recommended",  "may", and "optional" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+The keywords "must", "must not", "required", "shall", "shall not", "should", "should not",
+"recommended",  "may", and "optional" in this document are to be interpreted as described in [RFC
+2119](https://tools.ietf.org/html/rfc2119).
 
-## Prerequisites
-Currently, Data Lake allows read/write access for all users authenticated via `DATALAKE_SAML_IDENTITY_PROVIDER_ARN` during deployment time (see [README](README.md#aws-infrastructure-deployment-cdk-stack)).
-
-The product team will provide you the following information to enable to you to start using the Data Lake:
-
-Also, following information must be provided by Data Lake instance maintainer in order to star using it:
+# Prerequisites
+## Data Lake account and resource names
+Following information must be provided by Data Lake instance maintainer in order to start using it:
 
 - Data Lake AWS account ID (`DATALAKE_AWS_ACCOUNT_ID`)
 - Data Lake user role name (`DATALAKE_USER_ROLE_NAME`)
 - Data Lake lambda function endpoint names (`DATALAKE_LAMBDA_FUNCTION_ENDPOINT_NAME`)
 
-### Data Maintainers
-To import data in to the Data Lake, you will need a 'staging' S3 bucket with your data in it. You will need to give permissions to Data Lake AWS account to read your data.
+## Dataset source S3 bucket
+To import data in to the Data Lake, dataset source S3 bucket must be readable by
+Data Lake.
 
-Example bucket policy:
-
+Example dataset source S3 bucket policy:
 ```
 {
     "Version": "2012-10-17",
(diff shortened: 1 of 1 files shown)


USAGE.md
```diff
@@ -74,16 +61,44 @@ The GDL performs many checks on datasets. If any of the checks fail the dataset
...
+# Authentication and authorization
+Data Lake allows read/write access for users authorized by SAML identity provider
+(`DATALAKE_SAML_IDENTITY_PROVIDER_ARN`) configured during deployment time (see
+[README](README.md#aws-infrastructure-deployment-cdk-stack)).
+
+Example of AWS service account authentication and authorization in to Data Lake users role via
+Azure:
+
+- `~/.aws/config` file
+    ```
```
l0b0, line 74: List indentation in Markdown is three spaces.

USAGE.md
```diff
@@ -74,16 +61,44 @@ The GDL performs many checks on datasets. If any of the checks fail the dataset
...
+    azure_tenant_id=<AZURE_TENANT_ID>
+    azure_app_id_uri=https://signin.aws.amazon.com/saml\#<AZURE_APP_ID>
+    azure_default_username=<AZURE_USER_NAME>
+    azure_default_role_arn=arn:aws:iam::<DATALAKE_AWS_ACCOUNT_ID>:role/<DATALAKE_USER_ROLE_NAME>
+    azure_default_duration_hours=12
+    azure_default_remember_me=true
+    ```
+
+- Log in to Data Lake AWS account using Data Lake users role
+```bash
```
l0b0, line 85: This should be indented.

USAGE.md
```diff
@@ -74,16 +61,44 @@ The GDL performs many checks on datasets. If any of the checks fail the dataset
...
+
+# Data Lake endpoints
 
 There are two end user interaction points in GDL:
 
 - A [dataset space endpoint](TODO), to create, get, update or delete individual datasets, and to list all datasets
 - A [dataset versions endpoint](TODO), to create new versions of datasets. The S3 files which constitute the dataset are all linked to a specific dataset version.
 
-Both of these are implemented as AWS Lambda functions, which means they can be run ("invoked") either via the AWS web interface (links above) or via any tool using the AWS API, such as the commands below.
+Both of these are implemented as AWS Lambda functions, which means they can be run ("invoked")
```
l0b0, line 97: Can you share your Markdown line split configuration or disable it? We keep writing long lines and then splitting them, which makes the diffs unnecessarily big.

USAGE.md
```diff
@@ -74,16 +61,44 @@ The GDL performs many checks on datasets. If any of the checks fail the dataset
...
+    azure_default_remember_me=true
+    ```
+
+- Log in to Data Lake AWS account using Data Lake users role
+```bash
+aws-azure-login --profile data-lake-users
+```
+
+
+# Data Lake endpoints
```
l0b0, line 90: This can just be "Endpoints".

USAGE.md
```diff
@@ -1,24 +1,25 @@
...
 - Data Lake lambda function endpoint names (`DATALAKE_LAMBDA_FUNCTION_ENDPOINT_NAME`)
 
-### Data Maintainers
-To import data in to the Data Lake, you will need a 'staging' S3 bucket with your data in it. You will need to give permissions to Data Lake AWS account to read your data.
+## Staging bucket permissions
+To import data in to the Data Lake, Staging S3 Bucket (dataset source bucket) must be readable by
+Data Lake.
 
-Example bucket policy:
+Example Staging Bucket policy:
```
l0b0, line 22: We have a *specific* bucket called the "staging bucket", but this configuration would work with any source bucket. Would it make sense to call it the "source bucket" here?

l0b0 (changes requested):

USAGE.md
```diff
@@ -74,16 +61,44 @@ The GDL performs many checks on datasets. If any of the checks fail the dataset
...
+Example of AWS service account authentication and authorization in to Data Lake users role via
+Azure:
+
+- `~/.aws/config` file
+    ```
+    [profile data-lake-users]
+    azure_tenant_id=<AZURE_TENANT_ID>
+    azure_app_id_uri=https://signin.aws.amazon.com/saml\#<AZURE_APP_ID>
+    azure_default_username=<AZURE_USER_NAME>
+    azure_default_role_arn=arn:aws:iam::<DATALAKE_AWS_ACCOUNT_ID>:role/<DATALAKE_USER_ROLE_NAME>
```
billgeo, line 79: This script  auto populates the aws config from https://myapplications.microsoft.com/
https://confluence.linz.govt.nz/display/GEOD/Login+to+AWS+Service+Accounts+via+Azure+in+Command+Line

So could maybe change the usage guidance could mirror that to avoid confusion?

Something like:

```
[profile li-datalake-prod]
azure_tenant_id=<AZURE_TENANT_ID>
azure_app_id_uri=<AZURE_APP_ID>
azure_default_username=<AZURE_USER_NAME>
azure_default_role_arn=
azure_default_remember_me=true
azure_default_duration_hours=12
```

And then 

```bash
aws-azure-login --profile li-datalake-prod
```

billgeo (dismissed):
One suggested change to align it a bit more with how users will authenticate. But otherwise LGTM.

USAGE.md
billgeo: It means they'll have to select which role they want in the CLI (if they have more than one), but probably not a big issue.

USAGE.md
imincik: I don't think keeping those long lines is a good idea. I would make equally short as our source code is.

USAGE.md
imincik: Indentation was fixed

USAGE.md
imincik: Indentation was fixed

USAGE.md
imincik: Changed to Endpoints

USAGE.md
imincik: Changed to "dataset source bucket"

USAGE.md
imincik: @billgeo , even after reading your comments multiple times I don't understand what do you mean. Do you want me just keep empty value for `azure_default_role_arn` ?

billgeo: Yes, I just have an empty value for `azure_default_role_arn` and it works fine.

I followed the instructions in that confluence page and it auto populates the aws config file with the value empty. In practice, this is what most LINZ users will be doing I think, so I thought it would be good to have this usage example the same as what that confluence page script does.

USAGE.md
imincik: I agree to have something formatting lines. For now I keep it as is and we should try to fix it. I was trying to use prettier, but looks like it needs some configuration.

USAGE.md
imincik: Instructions updated to what we have in Confluence, thanks

Issue #453, opened 2021-03-24
Metadata validation should continue after encountering multiple directories error


l0b0:
### User Story

So that I can get as much error information as possible when importing, as a dataset maintainer, I want to know about all "multiple directories" errors instead of only the first one.

#### Acceptance Criteria

- [ ] Given a dataset with more than one metadata file in a different directory from the root metadata file, when importing the dataset, then all metadata files in other directories are reported as user errors.
- [ ] Given a dataset with more than one asset file in a different directory from the root metadata file, when importing the dataset, then all asset files in other directories are reported as user errors.

#### Tasks

- [ ] ...
- [ ] ...

#### Definition of Ready
- [ ] This story is __ready__ to work on, according to the [team's definition](https://confluence.linz.govt.nz/pages/viewpage.action?pageId=87930423)

#### Definition of Done
- [ ] This story is __done__, according to the [team's definition](https://confluence.linz.govt.nz/pages/viewpage.action?pageId=87930423)

PR #459, opened 2021-03-25
feat: validate all same directory checks at once

MitchellPaff:
## Issues

https://github.com/linz/geospatial-data-lake/issues/453

### DIFF
--- a/backend/check_stac_metadata/utils.py
+++ b/backend/check_stac_metadata/utils.py
@@ -100,23 +100,11 @@ class STACDatasetValidator:
             )
             raise
         self.validation_result_factory.save(url, Check.JSON_SCHEMA, ValidationResult.PASSED)
-
-        url_prefix = get_url_before_filename(url)
-
         self.dataset_metadata.append({"url": url})
 
         for asset in object_json.get("assets", {}).values():
             asset_url = asset["href"]
-            asset_url_prefix = get_url_before_filename(asset_url)
-
-            if url_prefix != asset_url_prefix:
-                error_message = f"“{url}” links to asset file in different directory: “{asset_url}”"
-                self.validation_result_factory.save(
-                    url,
-                    Check.MULTIPLE_DIRECTORIES,
-                    ValidationResult.FAILED,
-                    details={"message": error_message},
-                )
+            self.validate_directory(asset_url, url)
 
             asset_dict = {"url": asset_url, "multihash": asset["checksum:multihash"]}
             self.logger.debug(dumps({"asset": asset_dict}))
@@ -125,19 +113,7 @@ class STACDatasetValidator:
         for link_object in object_json["links"]:
             next_url = link_object["href"]
             if next_url not in self.traversed_urls:
-                next_url_prefix = get_url_before_filename(next_url)
-
-                if self.traversed_urls[0] != next_url_prefix:
-                    error_message = (
-                        f"“{next_url}” exists in a different directory"
-                        f" to the root metadata file: “{self.traversed_urls[0]}”"
-                    )
-                    self.validation_result_factory.save(
-                        url,
-                        Check.MULTIPLE_DIRECTORIES,
-                        ValidationResult.FAILED,
-                        details={"message": error_message},
-                    )
+                self.validate_directory(next_url, url)
                 self.validate(next_url)
 
     def get_object(self, url: str) -> Any:
@@ -168,6 +144,20 @@ class STACDatasetValidator:
                 url=asset["url"],
                 multihash=asset["multihash"],
             ).save()
+
+    def validate_directory(self, url: str, parent_metadata_url: str) -> None:
+        root_path = get_url_before_filename(self.traversed_urls[0])
+        if root_path != get_url_before_filename(url):
+            self.validation_result_factory.save(
+                url,
+                Check.MULTIPLE_DIRECTORIES,
+                ValidationResult.FAILED,
+                details={
+                    "message": f"“metadata file: {parent_metadata_url} links to {url}”"
+                    f" which exists in a different directory to the root "
+                    f"metadata file directory: “{root_path}”"
+                },
+            )
 
 
 def get_url_before_filename(url: str) -> str:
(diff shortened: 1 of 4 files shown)


tests/test_check_stac_metadata.py
```diff
@@ -106,6 +105,7 @@ def should_save_multiple_directories_validation_results(
     version_id = any_dataset_version_id()
     base_url = f"s3://{staging_bucket_name}/"
     invalid_key = f"test/{any_safe_filename()}"
+    valid_key = any_safe_filename()
```
l0b0, line 108: We should probably make *both* asset keys invalid, and then assert that we save *two* multiple directories errors.

tests/test_check_stac_metadata.py
```diff
@@ -106,6 +105,7 @@ def should_save_multiple_directories_validation_results(
     version_id = any_dataset_version_id()
     base_url = f"s3://{staging_bucket_name}/"
     invalid_key = f"test/{any_safe_filename()}"
```
l0b0, line 107: We might want to use `any_safe_filename()` instead of "test" since the actual name of the directory doesn't matter. Ditto in the other test.

tests/test_check_stac_metadata.py
```diff
@@ -159,6 +168,73 @@ def should_save_multiple_directories_validation_results(
...
 
+@mark.infrastructure
+@patch("backend.check_stac_metadata.task.ValidationResultFactory")
+def should_save_metadata_multiple_directories_validation_results(
+    validation_results_factory_mock: MagicMock,
+    subtests: SubTests,
+) -> None:
+    staging_bucket_name = get_param(ParameterName.STAGING_BUCKET_NAME)
+    base_url = f"s3://{staging_bucket_name}/"
+    invalid_child_key = f"test/{any_safe_filename()}"
```
l0b0, line 179: We should probably have a grandchild metadata file and assert that we get two errors. That way we can strengthen the assertions:

- We continue parsing metadata even in other directories, so that the user gets as much information as possible
- The last error message mentions the *root* URL even when parsing the child of a metadata file in another directory (this might require changing `url_prefix != next_url_prefix` and the accompanying error message to instead use `self.traversed_urls[0]` as the reference URL)

l0b0 (changes requested):

tests/test_check_stac_metadata.py
```diff
@@ -134,7 +143,7 @@ def should_save_multiple_directories_validation_results(
         ]
 
         with subtests.test(msg="Exit code"):
-            assert main() == 1
+            assert main() == 0
 
         hash_key = f"DATASET#{dataset_id}#VERSION#{version_id}"
         expected_error = "“{}” links to asset file in different directory: “{}”".format(
```
l0b0, line 149: There should be more errors after introducing another wrong key.

tests/test_check_stac_metadata.py
```diff
@@ -159,6 +168,103 @@ def should_save_multiple_directories_validation_results(
...
+                        Check.JSON_SCHEMA,
+                        ValidationResult.PASSED,
+                    ),
+                    call().save(
+                        invalid_child_s3_object.url,
+                        Check.MULTIPLE_DIRECTORIES,
+                        ValidationResult.FAILED,
+                        details={
+                            "message": expected_error.format(
+                                invalid_child_s3_object.url, invalid_grandchild_s3_object.url
```
l0b0, line 255: As discussed, the error message should refer to the *root* rather than the *parent* URL. For example, if we have root key `s3://bucket/foo.json`, child `s3://bucket/dir/bar.json` and grandchild `s3://bucket/baz.json`, the grandchild is in the *same* directory as the root and should not provoke an error.

l0b0 (changes requested):

backend/check_stac_metadata/utils.py
```diff
@@ -117,19 +117,27 @@ def validate(self, url: str) -> None:  # pylint: disable=too-complex
...
-        for link_object in url_json["links"]:
+        for link_object in object_json["links"]:
             next_url = link_object["href"]
             if next_url not in self.traversed_urls:
                 next_url_prefix = get_url_before_filename(next_url)
-                assert (
-                    url_prefix == next_url_prefix
-                ), f"“{url}” links to metadata file in different directory: “{next_url}”"
+
+                if self.traversed_urls[0] != next_url_prefix:
```
l0b0, line 130: As discussed, this isn't doing the right comparison (a full URL against a prefix). Since all the tests are passing we have a hole in the tests.

backend/check_stac_metadata/utils.py
```diff
@@ -117,19 +117,27 @@ def validate(self, url: str) -> None:  # pylint: disable=too-complex
...
+                    error_message = (
+                        f"“{next_url}” exists in a different directory"
```
l0b0, line 132: As discussed, this and the asset URL error message should be the same. The code should probably also be pulled out since it can be made identical for assets and metadata URLs.

l0b0 (changes requested):

backend/check_stac_metadata/utils.py
```diff
@@ -161,6 +145,20 @@ def save(self, key: str) -> None:
...
 
+    def validate_directory(self, url: str, parent_metadata_url: str) -> None:
+        root_path = get_url_before_filename(self.traversed_urls[0])
+        if root_path != get_url_before_filename(url):
+            self.validation_result_factory.save(
+                url,
+                Check.MULTIPLE_DIRECTORIES,
+                ValidationResult.FAILED,
+                details={
+                    "message": f"“metadata file: {parent_metadata_url} links to {url}”"
```
l0b0, line 156: - I suspect the typographic quotes should be around the two URLs, not this whole line.
- Do we want to generally capitalize the first word in these error messages?
- Should there be a colon in this line?

backend/check_stac_metadata/utils.py
```diff
@@ -161,6 +145,20 @@ def save(self, key: str) -> None:
...
+                    f" which exists in a different directory to the root "
+                    f"metadata file directory: “{root_path}”"
+                },
+            )
+
 
 def get_url_before_filename(url: str) -> str:
```
l0b0, line 163: We should `@lru_cache` this endpoint since it now gets run many times for the root URL.

l0b0 (changes requested):

l0b0 (approved):

Issue #465, opened 2021-03-26
Improve file validation performance


billgeo:
### User Story

So that I can work efficiently, as a Data Maintainer, I want the system to validate my data as quickly as possible

Note: in a test on 26/03/2021, validated and imported ~40GB of data, ~130 file at 300-550MB each. It took 46 minutes in total, most of which seemed to be doing the file checksums. User expectation is probably no more than 10-15 mins for validation for a dataset of this nature.

See here for the test on 26/03:
https://ap-southeast-2.console.aws.amazon.com/states/home?region=ap-southeast-2#/executions/details/arn:aws:states:ap-southeast-2:632223577832:execution:nonproddatasetversioncreation8597518E-ssEJMVu9eSjV:a7a295378dce11ebb9a4a57730b731ed

#### Acceptance Criteria

- [ ] Given [a pre-condition], when [an event happens], then [an expected outcome occurs]
- [ ] Given [a pre-condition], when [an event happens], then [an expected outcome occurs]

#### Tasks

- [ ] ...
- [ ] ...

#### Definition of Ready
- [ ] This story is __ready__ to work on, according to the [team's definition](https://confluence.linz.govt.nz/pages/viewpage.action?pageId=87930423)

#### Definition of Done
- [ ] This story is __done__, according to the [team's definition](https://confluence.linz.govt.nz/pages/viewpage.action?pageId=87930423)

PR #474, opened 2021-03-29
feat: increase batch array size to maximum amount

MitchellPaff:
## Issues

https://github.com/linz/geospatial-data-lake/issues/465

### DIFF
--- a/backend/content_iterator/task.py
+++ b/backend/content_iterator/task.py
@@ -3,10 +3,7 @@ from jsonschema import validate  # type: ignore[import]
 from ..processing_assets_model import ProcessingAssetType, ProcessingAssetsModel
 from ..types import JsonObject
 
-# From https://docs.aws.amazon.com/batch/latest/userguide/array_jobs.html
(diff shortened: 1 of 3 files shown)


backend/content_iterator/task.py
```diff
@@ -3,10 +3,7 @@
 from ..processing_assets_model import ProcessingAssetType, ProcessingAssetsModel
 from ..types import JsonObject
 
-# From https://docs.aws.amazon.com/batch/latest/userguide/array_jobs.html
-# TODO: Set MAX_ITERATION_SIZE to 10_000 once we figure out [how to set a numeric
-# size](https://stackoverflow.com/q/66202138/96588)
-MAX_ITERATION_SIZE = 5
+MAX_ITERATION_SIZE = 10000
```
l0b0, line 6: Looking at where this is used:

- `any_item_index` should be renamed to `any_next_item_index`
- `should_return_zero_as_first_item_if_no_content` should use `any_item_count` instead of `any_item_index`

l0b0 (approved):

Issue #336, opened 2021-03-04
Support for catalog/item STAC files


l0b0:
### User Story

So that we can import more diverse datasets, as a dataset maintainer, I *might* want to import a catalog or item STAC files in addition to the currently supported collections.

Datasets we may come across, in order of priority:

1. Collection -> Asset(s)
2. Collection -> Item(s) -> Asset(s)

#### Acceptance Criteria

- [ ] Given I have a STAC data set which contains a metadata file of type "Catalog" , when the data set is imported in to the data lake, then this metadata file is validated against the Catalog specification and no unexpected validations occur
- [ ] Given I have a STAC data set which contains a metadata file of type "Item" , when the data set is imported in to the data lake, then this metadata file is validated against the Item specification and no unexpected validations occur
- [ ] Given I have a STAC data set which contains a metadata file of type "Collection" , when the data set is imported in to the data lake, then this metadata file is validated against the Collection specification and no unexpected validations occur
- [ ] Given the above requirements, when a data set is imported in to the data lake, the time taken to process and validate the STAC metadata does not extend beyond reasonable (?!) processing times
- [ ] Given a catalog or collection or item with a reference to an asset, when importing the data set, then the item is copied across 

#### Additional context

STAC spec 1.0 has ways to unambiguously determine whether a file is a catalogue, collection or item. Unfortunately differentiating between a catalogue or collection [requires actually parsing the JSON file](https://github.com/radiantearth/stac-spec/blob/2c4ffa26691c1369598ccfdb8e1e6909bc782829/best-practices.md#how-to-differentiate-stac-files) before validating it, so it might be too slow:

> When crawling a STAC implementation, one can also make use of the relation type (rel field) when following a link. If it is an item rel type then the file must be a STAC Item. If it is child, parent or root then it must be a Catalog or a Collection, though the final determination between the two requires looking at the the type field in the Catalog or Collection JSON that is linked to.

#### Tasks

- [ ] Add minimal STAC objects to test framework
- [ ] update tests to use various different STAC types
- [ ] Test the performance of loading full dataset

#### Definition of Ready
- [ ] This story is __ready__ to work on, according to the [team's definition](https://confluence.linz.govt.nz/pages/viewpage.action?pageId=87930423)

#### Definition of Done
- [ ] This story is __done__, according to the [team's definition](https://confluence.linz.govt.nz/pages/viewpage.action?pageId=87930423)

billgeo:
Replaced by another story

PR #484, opened 2021-04-01
feat: support catalog and item metadata

MitchellPaff:
## Issues

https://github.com/linz/geospatial-data-lake/issues/336

### DIFF
--- a/tests/test_check_stac_metadata.py
+++ b/tests/test_check_stac_metadata.py
@@ -588,7 +588,7 @@ def should_raise_exception_if_non_s3_url_is_passed() -> None:
         STACDatasetValidator(url_reader, MockValidationResultFactory()).run(https_url)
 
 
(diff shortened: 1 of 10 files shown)


tests/test_check_stac_metadata.py
```diff
@@ -581,10 +581,48 @@ def should_raise_exception_if_non_s3_url_is_passed() -> None:
         STACDatasetValidator(url_reader, MockValidationResultFactory()).run(https_url)
 
 
-def should_return_assets_from_validated_metadata_files(subtests: SubTests) -> None:
+def should_return_assets_from_validated_collection_metadata_files(subtests: SubTests) -> None:
```
l0b0, line 591: I don't think "return" in the name is appropriate - we're asserting on a field of the `validator`. Ditto in the next test, I think.

l0b0 (approved):

l0b0 (approved):

PR #513, opened 2021-04-07
refactor: Modified stac validation task to not crash pipeline

MitchellPaff:


backend/check_stac_metadata/utils.py
```diff
@@ -74,16 +74,21 @@ def run(self, metadata_url: str) -> None:
...
-        self.traversed_urls.append(url)
         try:
-            object_json = self.get_object(url)
-        except JSONDecodeError:
+            self.validate(metadata_url)
+        except (ValidationError, ClientError, JSONDecodeError) as error:
+            LOGGER.error(dumps({"success": False, "message": str(error)}))
             return
 
+        self.save(hash_key)
```
l0b0, line 86: Do we want to inline the `save` method? It's only being called here, as far as I can tell.

backend/check_stac_metadata/task.py
```diff
@@ -31,16 +28,7 @@ def main() -> int:
...
-
-    except (AssertionError, ValidationError, ClientError) as error:
-        LOGGER.error(dumps({"success": False, "message": str(error)}))
-        return 1
-
-    validator.save(hash_key)
-
-    LOGGER.info(dumps({"success": True, "message": ""}))
+    validator.run(arguments.metadata_url, hash_key)
     return 0
```
l0b0, line 32: At this point we never actively return anything other than 0, so it's redundant. We can get rid of this line and simplify `sys.exit(main())` to just `main()`.

l0b0 (changes requested):

l0b0 (approved):

l0b0 (approved):

PR #519, opened 2021-04-08
refactor: Use key variables for common step function and error strings

MitchellPaff:


backend/import_dataset/task.py
```diff
@@ -47,14 +49,7 @@
 
 S3_BATCH_COPY_ROLE_ARN = get_param(ParameterName.IMPORT_DATASET_ROLE_ARN)
 
-DATASET_ID_KEY = "dataset_id"
-VERSION_ID_KEY = "version_id"
-METADATA_URL_KEY = "metadata_url"
-
 EVENT_KEY = "event"
```
l0b0, line 52: We can pull this out as a generic thing as well, based on its use case.

l0b0 (changes requested):

l0b0 (approved):

backend/import_dataset/task.py
l0b0: Cancelled - we can do this later.

PR #524, opened 2021-04-09
docs: Simplify usage

l0b0:
## Reference

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -97,7 +97,6 @@ export DATALAKE_LAMBDA_FUNCTION_ENDPOINT_NAME=<DATALAKE-LAMBDA-FUNCTION-ENDPOINT
 
 aws lambda invoke \
     --function-name "$DATALAKE_LAMBDA_FUNCTION_ENDPOINT_NAME" \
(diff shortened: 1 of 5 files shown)


MitchellPaff (approved):

billgeo (commented):
As far as the USAGE.md changes go, it LGTM. Will leave to other to approve the code changes.

imincik (commented):
I like this change.

The only small problem is env prefix in function name (for example `prod-datasets`). It might be confusing if Data Lake is deployed to environment with different name. So maybe better to set and use `ENV` variable here.

l0b0:
> The only small problem is env prefix in function name (for example `prod-datasets`). It might be confusing if Data Lake is deployed to environment with different name. So maybe better to set and use `ENV` variable here.

But we're going to deploy to a single prefix (or no prefix) in any case, right? And that production instance is probably going to run for years under the same name. What is the use case for end users using any of the other environments?

imincik:
This is public repo and users can deploy under any prefix and we have other test envs as well

imincik (dismissed):

imincik (approved):

PR #529, opened 2021-04-09
ci: production deployment on tag

imincik:
Deploy to Production environment if release tag ('release-*`) is created in release branch (`release-*). Report list of stack changes if change is pushed to release branch without tag.


.github/workflows/deploy.yml
```diff
@@ -1,4 +1,4 @@
-name: Deploy
+name: Deployment
```
l0b0, line 1: Why change this name? They are usually "Verb" or "Verb noun", like "Test" or "Deploy production".

.github/workflows/deploy.yml
```diff
@@ -87,10 +87,10 @@ jobs:
         DEPLOY_ENV: nonprod
         DATALAKE_SAML_IDENTITY_PROVIDER_ARN: ${{ secrets.DATALAKE_SAML_IDENTITY_PROVIDER_ARN_NON_PROD }}
 
-    # PROD DEPLOYMENT
+    # PROD DEPLOYMENT - in release branch, deploy on tag, otherwise report stack changes only
     - name: (Prod) Configure AWS credentials
       if: >
-        startsWith(github.ref, 'release')
+        startsWith(github.ref, 'refs/heads/release-') || startsWith(github.ref, 'refs/tags/release-')
```
l0b0, line 93: It would be simpler if we only deploy on tags, rather than allowing either. What do you think?

l0b0 (changes requested):

.github/workflows/deploy.yml
imincik: This is to make title consistent with other workflows - we have "Pull request" (which I don't like too much and I would recommend to rename it and "CodeQL Analysis"

.github/workflows/deploy.yml
imincik: It is only setting of credentials. I allow both actions  - push to branch and tag because if action is only push to release branch we report `cdk diff` with current prod deployment. Production deployment happens only if tag is pushed. But credentials are needed for both cases.

l0b0 (approved):

PR #533, opened 2021-04-12
feat(infra): add version tags to all AWS resources

imincik:
This change implements `Version` and `GitVersion` tags to all AWS resources in stack.

* `GitVersion`: always set to `<branch>/<commit>` value
* `Version`: if current commit is tag then set to `<tag>` otherwise set to `UNRELEASED`

## Reference


app.py
```diff
@@ -70,12 +71,45 @@ def main() -> None:
...
+    if not git_tag:
+        git_tag = "UNRELEASED"
+
     Tag.add(app, "CostCentre", "100005")
     Tag.add(app, APPLICATION_NAME_TAG_NAME, APPLICATION_NAME)
     Tag.add(app, "Owner", "Bill M. Nelson")
     Tag.add(app, "EnvironmentType", ENV)
     Tag.add(app, "SupportType", "Dev")
     Tag.add(app, "HoursOfOperation", "24x7")
+    Tag.add(app, "GitVersion", f"{git_branch}/{git_commit}")
```
l0b0, line 111: Could we split this up into branch + commit separately? Just in case we ever need to use the value in code, so we don't need to parse it.

imincik: Git information split in to `GitBranch` and `GitVersion` tags.

PR #534, opened 2021-04-12
ci: production AWS stack upgrade testing before deployment

imincik:
## Reference

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,200 @@ on:
     - release-*
 
 jobs:
+  test-prod-copy-deploy:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python: ['3.8']
+
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v2
+      with:
+        submodules: 'true'
+        fetch-depth: 0
+
+    # prod version detection
+    - name: Configure production AWS credentials
+      if: >
+        startsWith(github.ref, 'refs/heads/release-') || startsWith(github.ref, 'refs/tags/release-')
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -8,7 +8,200 @@ on:
...
+      if: >
+        startsWith(github.ref, 'refs/heads/release-') || startsWith(github.ref, 'refs/tags/release-')
+        && github.repository == 'linz/geospatial-data-lake'
+      run: |
+        poetry run cdk bootstrap aws://unknown-account/ap-southeast-2
+        poetry run cdk deploy --all --require-approval never
+
+
+  upgrade-copy-of-prod:
+    needs: deploy-copy-of-prod
```
l0b0, line 116: Is this intentional? Doesn't this mean that we can't upgrade without running deploy *in the same pipeline* first?

.github/workflows/deploy.yml
```diff
@@ -8,7 +8,200 @@ on:
...
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python: ['3.8']
```
l0b0, line 120: There is so much duplicate code here now that I think we should probably pull out a few reusable snippets. What do you think?

l0b0 (changes requested):

.github/workflows/deploy.yml
imincik: Yes, there's a lot of duplication, but I would pull out code in separate PR

.github/workflows/deploy.yml
imincik: This is testing of prod upgrade done on prod copy in CI environment

Issue #43, opened 2020-10-26
Dataset Space: allow space deletion only if empty


imincik:
## User Story
So that I don't accidentally lose my important data, as Data Maintainer, I want to only delete datasets with no data in them.

Note: datasets can be altered in every way, so there should never be a need to delete them. We may implement an 'archive' feature later if needed.

## Accpetance Criteria
- Given a dataset with 1 or more versions, when a dataset DELETE is requested, then a message is returned and the dataset is not deleted
- Given a datasets with no versions created, when a dataset DELETE is requested, then the dataset is deleted and no longer appears in the list of datasets

## Tasks
- [ ]

## Ready
- [x] This story is ready

## Done
- [ ] This story is done.

billgeo:
Rewrite as user story

billgeo:
As discussed. This will be a problem, if we don't fix this I think.

billgeo:
hi @l0b0 I get the error described in #567 when tryin to test this. Can you please have a look.

PR #538, opened 2021-04-13
feat: Only allow deleting empty dataset

l0b0:
## Issues

Closes https://github.com/linz/geospatial-data-lake/issues/43.

## Reference

### DIFF
--- a/backend/datasets/delete.py
+++ b/backend/datasets/delete.py
@@ -1,12 +1,16 @@
 """Delete dataset function."""
 from http import HTTPStatus
 
+import boto3
 from jsonschema import ValidationError, validate  # type: ignore[import]
 from pynamodb.exceptions import DoesNotExist
 
 from ..api_responses import error_response, success_response
 from ..datasets_model import datasets_model_with_meta
+from ..parameter_store import ParameterName, get_param
 from ..types import JsonObject
+
+BOTO3_CLIENT = boto3.client("s3")
 
 
 def delete_dataset(payload: JsonObject) -> JsonObject:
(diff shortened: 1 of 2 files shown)


backend/datasets/delete.py
```diff
@@ -24,12 +26,23 @@ def delete_dataset(payload: JsonObject) -> JsonObject:
...
     try:
-        dataset = datasets_model_class.get(
-            hash_key=f"DATASET#{req_body['id']}", consistent_read=True
-        )
+        dataset = datasets_model_class.get(hash_key=f"DATASET#{dataset_id}", consistent_read=True)
     except DoesNotExist:
-        return error_response(HTTPStatus.NOT_FOUND, f"dataset '{req_body['id']}' does not exist")
+        return error_response(HTTPStatus.NOT_FOUND, f"dataset '{dataset_id}' does not exist")
+
+    # Verify that the dataset is empty
```
imincik, line 35: Did you consider getting information if dataset space is empty from DB as well ? I am not decided which approach - checking S3 bucket or DB is better.

l0b0: As far as I can tell there is nothing in the DB to check - we don't store any DB records related to dataset versions yet.

backend/datasets/delete.py
```diff
@@ -24,12 +26,23 @@ def delete_dataset(payload: JsonObject) -> JsonObject:
...
+    s3_client = boto3.client("s3")
```
MitchellPaff, line 36: not major but in other places we've initialised boto3 clients outside of the lambda handler.

MitchellPaff (approved):

PR #546, opened 2021-04-13
feat(infra): store deployment version as SSM parameter

imincik:
Store deployment version as SSM Parameter instead of CloudFormation stack tag. CloudFormation stack tags are applied recursively to all resources which is causing changes to all resources during all deployments and making CloudFormation changesets unreadable. This change is fixing this issue.

## Reference

### DIFF
--- a/infrastructure/storage_stack.py
+++ b/infrastructure/storage_stack.py
@@ -9,6 +9,7 @@ from aws_cdk.core import Construct, RemovalPolicy, Stack, Tags
 from backend.datasets_model import DatasetsTitleIdx
 from backend.environment import ENV
 from backend.parameter_store import ParameterName
+from backend.version import GIT_BRANCH, GIT_COMMIT, GIT_TAG
 
 from .constructs.table import Table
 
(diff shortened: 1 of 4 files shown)


infrastructure/storage_stack.py
```diff
@@ -23,6 +24,34 @@ def __init__(self, scope: Construct, stack_id: str, deploy_env: str, **kwargs: A
...
 
+        ############################################################################################
+        # ### DEPLOYMENT VERSION ###################################################################
+        ############################################################################################
+
+        aws_ssm.StringParameter(
+            self,
+            "git-branch",
+            parameter_name=f"/{ENV}/git_branch",
+            string_value=f"{GIT_BRANCH}",
```
l0b0, line 35: `f"{GIT_BRANCH}"` could be simplified to just `GIT_BRANCH`. Ditto elsewhere in this file.

l0b0 (changes requested):

infrastructure/storage_stack.py
imincik: Thank you, fixed now.

l0b0 (approved):

PR #569, opened 2021-04-19
fix: ssm throttling issues

MitchellPaff:

### DIFF
--- a/backend/validation_results_model.py
+++ b/backend/validation_results_model.py
@@ -78,10 +78,9 @@ class ValidationResultFactory:  # pylint:disable=too-few-public-methods
     def __init__(
         self,
         hash_key: str,
-        results_table_name: Optional[str] = None,
+        results_table_name: str,
     ):
-        if results_table_name is None:
-            results_table_name = get_param(ParameterName.VALIDATION_RESULTS_TABLE_NAME)
+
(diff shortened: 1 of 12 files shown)


backend/validation_results_model.py
```diff
@@ -41,39 +41,49 @@ class ValidationResultsModelBase(Model):
...
 
     return ValidationResultsModel
 
 
 class ValidationResultFactory:  # pylint:disable=too-few-public-methods
-    def __init__(self, hash_key: str):
+    def __init__(
+        self,
+        hash_key: str,
+        results_table_name: Optional[str] = None,
```
l0b0, line 81: As discussed, we probably want to make this non-optional, to avoid the different code paths in production code.

tests/aws_utils.py
```diff
@@ -74,6 +75,10 @@ def any_item_count() -> int:
     return randrange(3)
 
 
+def any_table_name() -> str:
+    return f"{random_string(5)}_{random_string(5)}_{random_string(5)}"
```
l0b0, line 79: Why not just `return random_string(15)`?

l0b0 (changes requested):

l0b0 (approved):

l0b0 (approved):

PR #575, opened 2021-04-19
feat: less verbose AWS Batch jobs and Lambda functions bundling

imincik:
This change will produce less output during AWS infrastructure deployment which will result, hopefully, in more visible stack changes output.

## Reference


backend/Dockerfile
```diff
@@ -1,26 +1,29 @@
 FROM ubuntu:20.04 as build
 
-RUN apt-get update \
-    && apt-get install --assume-yes --no-install-recommends curl python3-pip \
+RUN apt-get update --quiet --quiet \
+    && apt-get install --quiet --quiet --assume-yes --no-install-recommends \
+    curl python3-pip < /dev/null > /dev/null \
```
l0b0, line 5: `< /dev/null` just closes standard input for the `apt-get` command. Does it actually make it less verbose? Also, with `--quiet --quiet` we probably don't want `> /dev/null` as well, that should be redundant.

Ditto for the other `apt-get`.

backend/Dockerfile
```diff
@@ -1,26 +1,29 @@
...
     && rm -rf /var/lib/apt/lists/*
+# hadolint ignore=DL3059
```
l0b0, line 7: Weird that hadolint didn't complain about this before.

backend/Dockerfile
```diff
@@ -1,26 +1,29 @@
...
 RUN curl --location --show-error --silent --output get-poetry.py https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py \
     && echo '08336beb0091ab192adef2cedbaa3428dabfc8572e724d2aa7fc4a4922efb20a get-poetry.py' > get-poetry.py.sha256 \
     && sha256sum --check get-poetry.py.sha256 \
-    && python3 get-poetry.py \
+    && python3 get-poetry.py > /dev/null \
```
l0b0, line 11: I do want to know where poetry gets installed, so I'd rather not do this.

l0b0 (changes requested):

backend/Dockerfile
imincik: I had to upgrade my hadolint version. Apparently this check came with new version.

backend/Dockerfile
imincik: I am trying to make `cdk diff` or `cdk synth` output readable here. These messages are not a big issue, but interrupt CDK output. But if you really want, I can keep those messages.

l0b0: I'd rather keep these messages, yes.

imincik: I think, this is the output we are talking about: 
```
# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

It will add the `poetry` command to Poetry's bin directory, located at:

$HOME/.poetry/bin

This path will then be added to your `PATH` environment variable by
modifying the profile file located at:

$HOME/.profile

You can uninstall at any time by executing this script with the --uninstall option,
and these changes will be reverted.

Installing version: 1.1.6
  - Downloading poetry-1.1.6-linux.tar.gz (72.33MB)

Poetry (1.1.6) is installed now. Great!

To get started you need Poetry's bin directory ($HOME/.poetry/bin) in your `PATH`
environment variable. Next time you log in this will be done
automatically.

To configure your current shell run `source $HOME/.poetry/env```` 

How much useful is this ?

l0b0: It mentions where Poetry is installed, which is useful for debugging.

backend/Dockerfile
imincik: Yes, it makes it less verbose then I was able to do with `--quiet`.

imincik: `< /dev/null` removed but I left stdout redirection `> /dev/null`.

backend/Dockerfile
imincik: I think it is not very hard to guess even if you don't have it in log output. I would say having as much readable CDK output as possible is more important, because this is always different and you can easily miss something there.

l0b0 (approved):

PR #607, opened 2021-04-21
feat: implement separate configuration for resource removal policy

imincik:
This change add separate resources removal policy in order to be able to set it to `RETAIN` for CI deployments which are testing production deployment upgrades.

## Reference

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -21,7 +21,7 @@ jobs:
           submodules: 'true'
           fetch-depth: 0
 
-        # prod version detection
+      # prod version detection
       - name: Configure production AWS credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
@@ -38,7 +38,7 @@ jobs:
       - name: Checkout to current production version
         run: git checkout ${{ env.PROD_VERSION }}
 
-        # dependencies installation
+      # dependencies installation
       - name: Use Node for CDK deployment
         uses: actions/setup-node@v2.1.5
         with:
@@ -82,7 +82,7 @@ jobs:
       - name: Print CDK version
         run: poetry run cdk --version
 
-        # deployment
+      # deployment
       - name: Configure CI AWS credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yml
```diff
@@ -174,6 +175,7 @@ jobs:
       run: |
         DEPLOY_ENV="ci${GITHUB_RUN_ID}"
         echo "DEPLOY_ENV=$DEPLOY_ENV" | tee -a $GITHUB_ENV
+        echo "RESOURCE_REMOVAL_POLICY=RETAIN" | tee -a $GITHUB_ENV
```
l0b0, line 178: Am I misunderstanding something, or does this make the deployment different from the actual production deployment? Wouldn't that make it likely that `test-prod-copy-upgrade` passes but `deploy` might fail?

l0b0 (changes requested):

.github/workflows/deploy.yml
imincik: `RESOURCE_REMOVAL_POLICY=RETAIN` makes our upgrade testing deployment similar to prod. Until now, removal policy was set to `RETAIN` only for prod deployments.

backend/environment.py
```diff
@@ -1,3 +1,12 @@
 import os
 
+from aws_cdk.core import RemovalPolicy
+
 ENV = os.environ.get("DEPLOY_ENV", "test")
+RESOURCE_REMOVAL_POLICY = os.environ.get("RESOURCE_REMOVAL_POLICY", "DESTROY")
```
l0b0, line 6: We can inline this, because the value is only used in one place.

backend/environment.py
```diff
@@ -1,3 +1,12 @@
...
+
+if RESOURCE_REMOVAL_POLICY.upper() == "RETAIN":
+    REMOVAL_POLICY = RemovalPolicy.RETAIN
```
l0b0, line 9: We should probably move this to the infrastructure directory, since it's only used there.

l0b0 (changes requested):

.github/workflows/deploy.yml
l0b0: Where is it set to retain prod deployments? `git grep -i RETAIN` doesn't show anything prod-related AFAICT.

backend/environment.py
imincik: Inlined now.

backend/environment.py
imincik: Code moved to `infrastructure/removal_policy.py` file

PR #621, opened 2021-04-22
refactor: add stack name prefix to SSM parameter names

imincik:
This change should make stack updates more safer in case we migrate some SSM parameters between stacks. Should fix current prod deployment testing failure.

Based on #620 which needs to be merged first.

## Reference


backend/parameter_store.py
```diff
@@ -21,15 +21,15 @@ def _generate_next_value_(  # type: ignore[misc,override] # pylint:disable=no-se
...
-    IMPORT_DATASET_ROLE_ARN = auto()
+    STORAGE_DATASETS_TABLE_NAME = auto()
+    PROCESSING_DATASET_VERSION_CREATION_STEP_FUNCTION_ARN = auto()
+    PROCESSING_IMPORT_ASSET_FILE_FUNCTION_TASK_ARN = auto()
+    PROCESSING_IMPORT_METADATA_FILE_FUNCTION_TASK_ARN = auto()
+    PROCESSING_IMPORT_DATASET_ROLE_ARN = auto()
     PROCESSING_ASSETS_TABLE_NAME = auto()
     STAGING_BUCKET_NAME = auto()
     STORAGE_BUCKET_NAME = auto()
-    VALIDATION_RESULTS_TABLE_NAME = auto()
```
l0b0, line 32: Why is it helpful to know which stack a parameter belongs to?

imincik: Knowing where parameter belongs is not very important. The reason of adding those stack name prefixes is that we need to make sure that parameter name is changed if it moves between stacks to avoid deployment issues (avoid same parameter name conflict).

backend/parameter_store.py
```diff
@@ -21,15 +21,15 @@ def _generate_next_value_(  # type: ignore[misc,override] # pylint:disable=no-se
...
+    STORAGE_VALIDATION_RESULTS_TABLE_NAME = auto()
```
l0b0, line 32: Can you please order these alphabetically? Also, don't we need to add the prefix to the staging bucket name?

imincik: @l0b0 , sorted now.

l0b0 (approved):

PR #646, opened 2021-04-30
refactor(infra): use nested stacks with cross-stack references passed using parameters

imincik:
This change should fix our stack updating issues (should not use CF stack exports) and allows us to keep resources logically separated in stacks (with exception of `botocore` stack which is merged with `storage` stack). This change is based on usage of nested stacks and passing resource information between them using CF parameters.

### DIFF
--- a/app.py
+++ b/app.py
@@ -8,11 +8,11 @@ from os import environ
 from aws_cdk.core import App, Environment, Stack, Tag
 
 from backend.environment import ENV
-from infrastructure.api_stack import APIStack, APIStackProps
+from infrastructure.api_stack import APIStack
 from infrastructure.constructs.batch_job_queue import APPLICATION_NAME, APPLICATION_NAME_TAG_NAME
 from infrastructure.lambda_layers_stack import LambdaLayersStack
-from infrastructure.lds import LDSStack, LDSStackProps
-from infrastructure.processing_stack import ProcessingStack, ProcessingStackProps
+from infrastructure.lds import LDSStack
+from infrastructure.processing_stack import ProcessingStack
 from infrastructure.staging_stack import StagingStack
 from infrastructure.storage_stack import StorageStack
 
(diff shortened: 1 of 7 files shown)


imincik:
@l0b0 , @MitchellPaff . This PR is passing tests now. Would you please have a look ? Thanks

infrastructure/api_stack.py
```diff
@@ -15,22 +16,27 @@
 from .roles import MAX_SESSION_DURATION
 
 
-class APIStack(Stack):
+@dataclass
```
l0b0, line 19: `NestedStackProps` isn't a dataclass, so I'm not sure this is appropriate. For example, `@dataclass` will override the `__init__` of the parent class, by default, at which point this class is not really a `NestedStackProps` anymore.

infrastructure/staging_stack.py
```diff
@@ -1,13 +1,20 @@
...
-    def __init__(self, scope: Construct, stack_id: str, *, deploy_env: str, **kwargs: Any) -> None:
+class StagingStack(NestedStack):
+    def __init__(
+        self,
+        scope: Construct,
+        stack_id: str,
+        *,
+        deploy_env: str,
+        **kwargs: Any,
+    ) -> None:
```
l0b0, line 17: The `__init__` signature still fits on one line. Ditto elsewhere.

infrastructure/storage_stack.py
```diff
@@ -102,4 +110,14 @@ def __init__(self, scope: Construct, stack_id: str, *, deploy_env: str, **kwargs
             ),
         )
 
+        ############################################################################################
+        # ### LAMBDA LAYERS BOTOCORE ###############################################################
+        ############################################################################################
+        self.botocore = aws_lambda_python.PythonLayerVersion(
```
l0b0, line 116: Does this actually belong in the storage stack?

l0b0 (changes requested):

app.py
```diff
@@ -24,65 +23,57 @@ def main() -> None:
...
     )
 
-    StagingStack(
-        app,
-        "staging",
-        deploy_env=ENV,
-        env=environment,
-        stack_name=f"{ENV}-geospatial-data-lake-staging",
-    )
+    StagingStack(datalake, "staging", deploy_env=ENV)
```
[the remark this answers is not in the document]
imincik: `StagingStack` is nested stack which doesn't take `env` or `stack_name` parameter.

app.py
```diff
@@ -24,65 +23,57 @@ def main() -> None:
         account=environ["CDK_DEFAULT_ACCOUNT"], region=environ["CDK_DEFAULT_REGION"]
     )
 
+    datalake = Stack(app, "datalake", env=environment)
```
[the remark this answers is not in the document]
imincik: `datalake` is top parent stack. Yes, it looks like I should improve configuration of `datalake` stack.

app.py
```diff
@@ -24,65 +23,57 @@ def main() -> None:
...
-        env=environment,
-        stack_name=f"{ENV}-geospatial-data-lake-processing",
+        props=processing_props,
     )
 
-    APIStack(
-        app,
-        "api",
-        botocore_lambda_layer=lambda_layers.botocore,
+    # can't get pylint working with Python dataclass
```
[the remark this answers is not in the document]
imincik: I like it better to have this comment in code so everybody knows why pylint is disabled here.

infrastructure/api_stack.py
imincik: `@dataclass` decorator is set for `APIStackProps` class which is what I wanted to do and this class behaves as I would expect from dataclass

infrastructure/staging_stack.py
imincik: This reformatting was was done by black.

app.py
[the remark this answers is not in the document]
imincik: I just added stack_name with ENV prefix

```
-    datalake = Stack(app, "datalake", env=environment)                                                                                                                                                                                        
+    datalake = Stack(app, "datalake", env=environment, stack_name="{ENV}-geospatial-data-lake")  
```

infrastructure/storage_stack.py
imincik: I'll try if moving botocore back to it's own stack is going to work

imincik: `lambda_layers_stack` removed from `storage` and returned back to its own stack.

infrastructure/staging_stack.py
l0b0: If you remove the trailing comma it'll revert to the original code.

infrastructure/api_stack.py
l0b0: Could you please explain why though? Is it *necessary* to make this a dataclass?

app.py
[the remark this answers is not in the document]
l0b0: I think this is happening because the dataclass overrides the `__init__` method.

l0b0 (changes requested):

app.py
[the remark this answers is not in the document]
imincik: I managed to remove `NestedStacksProps` completely.

infrastructure/staging_stack.py
imincik: Done now.

imincik:
@l0b0 , I managed to remove usage of  `NestedStackProps` construct completely and keep the same CloudFormation result. Now this PR makes only very simple change from sibling stacks to nested stacks which removes need to produce CF Exports to build cross-stack references.

l0b0 (approved):

PR #658, opened 2021-05-04
feat: added read only role for all linz users

MitchellPaff:


backend/resources.py
```diff
@@ -8,3 +8,4 @@ class ResourceName(Enum):
     DATASET_VERSIONS_ENDPOINT_FUNCTION_NAME = f"{ENV}-dataset-versions"
     IMPORT_STATUS_ENDPOINT_FUNCTION_NAME = f"{ENV}-import-status"
     USERS_ROLE_NAME = f"{ENV}-data-lake-users"
+    LINZ_READ_ONLY_ROLE_NAME = f"{ENV}-data-lake-linz-readonly"
```
l0b0, line 11: Would removing the "LINZ" part of the name make sense? We might want to grant access to it for non-LINZ users in the future.

l0b0 (changes requested):

l0b0 (dismissed):

l0b0 (approved):

PR #664, opened 2021-05-05
feat: Track generated CDK context

l0b0:
As per
<https://aws.amazon.com/blogs/devops/best-practices-for-developing-cloud-applications-with-aws-cdk/>.

This partially reverts commit e118a393626e7436a11257ea2a91f445c3b3a36a.

## Reference

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-cdk.context.json
 cdk.out/
 .coverage
 .idea/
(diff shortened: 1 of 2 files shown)


MitchellPaff (approved):

imincik:
It applies best practices but our `cdk.context.json` is not applicable for general public users, so it is questionable.

l0b0:
> It applies best practices but our `cdk.context.json` is not applicable for general public users, so it is questionable.

General public users will have to create their own VPC. I'm not sure whether that will result in regenerating cdk.context.json, but if not we might want to add a note to the documentation about removing it.

PR #668, opened 2021-05-06
feat: Bundle Lambda code outside Docker

l0b0:
This way the asset hash is generated from the actual asset code rather
than from the entire project directory.

## Reference

### DIFF
--- a/backend/bundle.bash
+++ b/backend/bundle.bash
@@ -2,11 +2,11 @@
 
 set -o errexit -o noclobber -o nounset
 
-if [[ "$#" -lt 1 ]]
+if [[ "$#" -ne 2 ]]
 then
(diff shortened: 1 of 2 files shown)


l0b0:
Looks like it's futile to avoid Lambdas rebuilding all the time, because Pip installs are not reproducible.￼ I was able to get most of the way there by adding `--no-compile` (avoids creating `__pycache__` directories) and removing `RECORD` files which contain some changing checksums, but even then the shebang lines in the `bin` directory point to the absolute directory path (meaning we'd have to build in the same directory every time, which would be OK but not great), and finally `pvectorc.cpython-38-x86_64-linux-gnu.so` is different between every build.

PR #669, opened 2021-05-06
refactor(infra): re-introduce users stack (closes #619)

### LINKED ISSUE #619: Move user roles back to separate users stack
Move "users-role" back to separate users stack which was changed in 0c44ffb6d32aedf626719a9547a0cfea189ada41 . This will allow us to exclude this role from deleting (and breaking Azure AD configuration) when re-deploying stacks

imincik:
## Reference


imincik:
@l0b0 , I can't find where the problem is. I can't see any parameter coming as constructor ID.

imincik:
[this](https://github.com/linz/geospatial-data-lake/blob/9bc9bd2a8f7ac8214d35effc936b774e78dd2f7a/infrastructure/constructs/lambda_endpoint.py#L34) is problematic like. If this one is disabled, then template is built without error.

Strange thing is that [other users role](https://github.com/linz/geospatial-data-lake/blob/9bc9bd2a8f7ac8214d35effc936b774e78dd2f7a/infrastructure/api_stack.py#L102) is not causing any problems.

imincik:
I am closing this PR as not working.

PR #676, opened 2021-05-06
feat: use static name for s3 buckets

MitchellPaff:


infrastructure/staging_stack.py
```diff
@@ -1,13 +1,13 @@
...
+from aws_cdk import aws_s3
 from aws_cdk.core import Construct, NestedStack, RemovalPolicy, Tags
 
-from backend.parameter_store import ParameterName
+from backend.resources import ResourceName
 
 
 class StagingStack(NestedStack):
-    def __init__(self, scope: Construct, stack_id: str, *, deploy_env: str, **kwargs: Any) -> None:
+    def __init__(self, scope: Construct, stack_id: str, **kwargs: Any) -> None:
```
l0b0, line 10: I think we can also remove `**kwargs`, since we never actually pass any extra parameters.

l0b0 (approved):

PR #677, opened 2021-05-07
refactor: removed kwargs argument from stacks

MitchellPaff:


infrastructure/lds_stack.py
```diff
@@ -14,9 +12,8 @@ def __init__(
         *,
         deploy_env: str,
         storage_bucket: aws_s3.Bucket,
-        **kwargs: Any,
```
l0b0, line 17: Does this signature fit on one line now? Ditto elsewhere.

l0b0 (approved):

PR #691, opened 2021-05-11
feat: dataset catalog

MitchellPaff:


backend/pystac_io_methods.py
```diff
@@ -0,0 +1,28 @@
...
+import boto3
+from pystac import STAC_IO  # type: ignore[import]
+
+
+def read_method(uri: str) -> Any:
+    parsed = urlparse(uri)
+    if parsed.scheme == "s3":
+        bucket = parsed.netloc
+        key = parsed.path[1:]
+        s3_resource = boto3.resource("s3")
```
l0b0, line 13: Is this a resource we could instantiate outside of the function to make sure it's initialized at import time?

backend/pystac_io_methods.py
```diff
@@ -0,0 +1,28 @@
...
+        obj = s3_resource.Object(bucket, key)
+        return obj.get()["Body"].read().decode("utf-8")
```
l0b0, line 15: We've used `S3_CLIENT.get_object(Bucket=bucket, Key=key)["Body"]` to get the equivalent elsewhere. We should probably use the same in both places, but I don't know whether there is any appreciable difference between these options.

poetry.lock
```diff
@@ -1145,288 +1145,290 @@ s3transfer = ">=0.3.0,<0.4.0"
...
-version = "1.17.61.0"
-description = "Type annotations for boto3 1.17.61, generated by mypy-boto3-buider 4.7.0"
+version = "1.17.71"
+description = "Type annotations for boto3 1.17.71, generated by mypy-boto3-buider 4.12.0"
 category = "dev"
 optional = false
 python-versions = ">=3.6"
 
 [package.dependencies]
-mypy-boto3-batch = {version = "1.17.61.0", optional = true, markers = "extra == \"batch\""}
```
l0b0, line 1155: I don't think we want to upgrade these packages as part of this change. One "simple" way to work around this would be to

0. Revert back to before this change
1. `poetry lock --no-update`
2. `git commit -m 'chore: Throw-away commit'`
3. `poetry add whatever`
4. `git commit`
5. `git rebase` and remove the throw-away commit

pyproject.toml
```diff
@@ -119,6 +120,8 @@ datasets = [
     "jsonschema",
     "pynamodb",
     "ulid-py",
+    "pystac",
+    "boto3",
```
l0b0, line 124: Could you please order the dependencies alphabetically?

tests/test_datasets_endpoint.py
```diff
@@ -2,50 +2,90 @@
...
+
+        catalog = get_s3_prefix_versions(
+            ResourceName.STORAGE_BUCKET_NAME.value, dataset_title, s3_client
+        )[0]
+
+        with smart_open(
+            f"s3://{ResourceName.STORAGE_BUCKET_NAME.value}/{catalog['Key']}"
+        ) as new_root_metadata_file:
+
+            catalog_json = loads(new_root_metadata_file.read().decode("utf-8"))
```
l0b0, line 70: `loads(new_root_metadata_file.read().decode("utf-8"))` I think can be simplified to `load(new_root_metadata_file)`.

tests/test_datasets_endpoint.py
```diff
@@ -2,50 +2,90 @@
...
+
+            with subtests.test(msg="catalog title"):
+                assert catalog_json["title"] == dataset_title
+
+            with subtests.test(msg="catalog description"):
+                assert catalog_json["description"] == dataset_description
+
+            with subtests.test(msg="catalog valid"):
+                validate_dict(catalog_json)
```
l0b0, line 79: Since you're using pystac to create the catalogue we probably don't need to test it here.

l0b0 (changes requested):

backend/pystac_io_methods.py
```diff
@@ -0,0 +1,28 @@
+from typing import IO, Any, Union
+from urllib.parse import urlparse
+
+import boto3
+from pystac import STAC_IO  # type: ignore[import]
+
+S3_CLIENT = boto3.client("s3")
+
+
+def read_method(uri: str) -> Any:
```
l0b0, line 10: Can we be more specific than `Any` here?

l0b0 (dismissed):

l0b0 (approved):

PR #694, opened 2021-05-11
refactor(infra): rename S3 and API user roles

imincik:
## Reference

### DIFF
--- a/backend/resources.py
+++ b/backend/resources.py
@@ -4,10 +4,10 @@ from .environment import ENV
 
 
 class ResourceName(Enum):
+    API_USERS_ROLE_NAME = f"{ENV}-api-users"
     DATASETS_ENDPOINT_FUNCTION_NAME = f"{ENV}-datasets"
     DATASET_VERSIONS_ENDPOINT_FUNCTION_NAME = f"{ENV}-dataset-versions"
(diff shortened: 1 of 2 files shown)


backend/resources.py
```diff
@@ -4,10 +4,10 @@
 
 
 class ResourceName(Enum):
+    API_USERS_ROLE_NAME = f"{ENV}-api-users"
     DATASETS_ENDPOINT_FUNCTION_NAME = f"{ENV}-datasets"
     DATASET_VERSIONS_ENDPOINT_FUNCTION_NAME = f"{ENV}-dataset-versions"
     IMPORT_STATUS_ENDPOINT_FUNCTION_NAME = f"{ENV}-import-status"
-    READ_ONLY_ROLE_NAME = f"{ENV}-data-lake-readonly"
+    S3_USERS_ROLE_NAME = f"{ENV}-s3-users"
```
l0b0, line 11: I'd probably avoid using "S3" in the name, since we might add access to other resources. How about using a term like "consumers" or "read-only" to make it service-agnostic?

l0b0 (changes requested):

backend/resources.py
imincik: It was called "read-only". I usually try to avoid using technology name in such places but I think this is good place to use it. This user access is really determined by S3 features and API.

l0b0 (approved):

PR #722, opened 2021-05-18
docs: Manage renaming datasets

l0b0:
## Issues

Closes https://github.com/linz/geostore/issues/611.

## Reference

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -122,64 +122,98 @@ aws lambda invoke --function-name GEOSTORE-LAMBDA-FUNCTION-ENDPOINT-NAME \
 
 ## Dataset Space Endpoint Usage Examples
 
-- Example of Dataset creation request
+### Dataset creation request
 
-  ```console
-  $ aws lambda invoke --function-name datasets \
-      --payload '{"http_method": "POST", "body": {"title": "Auckland_2020"}}' /dev/stdout
+```console
+$ aws lambda invoke --function-name datasets \
+    --payload '{"http_method": "POST", "body": {"title": "Auckland_2020"}}' /dev/stdout
 
-  {"status_code": 201, "body": {"created_at": "2021-02-01T13:38:40.776333+0000", "id": "cb8a197e649211eb955843c1de66417d", "title": "Auckland_2020", "updated_at": "2021-02-01T13:39:36.556583+0000"}}
-  ```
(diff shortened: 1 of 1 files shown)


USAGE.md
```diff
@@ -122,72 +122,104 @@ aws lambda invoke \
...
+    --function-name "${ENV}-datasets" \
+    --payload '{"http_method": "DELETE", "body": {"id": "cb8a197e649211eb955843c1de66417d"}}' \
+    /dev/stdout
+
+{"status_code": 204, "body": {}}
+```
+
+### Changing the dataset title by creating a copy of the latest dataset version
+
+This is the simplest way to change a dataset title:
```
billgeo, line 178: Do we need to add in the downside of this. Like losing previous versions of the dataset?

MitchellPaff (dismissed):

USAGE.md
```diff
@@ -122,72 +122,104 @@ aws lambda invoke \
...
+
+This process is more cumbersome and time-consuming than the above, and results links to the old
+dataset being broken rather than slowly phased out, but can be followed if necessary.
+
+1. Send a request to the product team asking for a rename, specifying the current and new title of
+   the dataset. The product team then takes care of the rest of the process:
+   1. Schedule necessary downtime and notify users.
+   1. Turn off external access to the whole or part of the system to avoid any conflicts.
+   1. Run a manual rename of all the files in the relevant dataset.
+   1. Use the `${ENV}-datasets` endpoint to rename the dataset in the database.
```
billgeo, line 197: Does 'renaming' a dataset exist or work on the datasets endpoint yet? There's nothing further down in the docs about it?

billgeo (changes requested):

USAGE.md
l0b0: The functionality is on the "datasets" endpoint, but we've left it out of the usage documentation since it's not meant to be used by end users. TBC whether we want to just remove it from public view, since I guess it'd be a fair bit of work to make that particular functionality accessible only to admins.

USAGE.md
l0b0: I've mentioned further down that asking to remove the existing dataset is an optional step. I don't expect we can just unconditionally get rid of the original dataset.

l0b0: I've documented how the datasets have no explicit connection between them.

USAGE.md
billgeo: Sounds good to me.

billgeo (approved):
Thanks @l0b0

PR #727, opened 2021-05-18
feat: root catalog

MitchellPaff:


backend/stac_format.py
```diff
@@ -10,12 +10,15 @@
...
-STAC_LINKS_KEY = "links"
 STAC_PROPERTIES_DATETIME_KEY = "datetime"
 STAC_PROPERTIES_KEY = "properties"
 STAC_TITLE_KEY = "title"
 STAC_TYPE_KEY = "type"
 STAC_VERSION_KEY = "stac_version"
+STAC_LINKS_KEY = "links"
+STAC_REL_KEY = "rel"
+STAC_ROOT_KEY = "root"
+STAC_CHILD_KEY = "child"
```
l0b0, line 21: "root" and "child" aren't keys. Maybe `STAC_REL_ROOT` or `STAC_REL_ROOT_VALUE`?

backend/write_to_catalog/task.py
```diff
@@ -0,0 +1,57 @@
...
+from ..types import JsonObject
+
+STAC_IO.write_text_method = write_method
+STAC_IO.read_text_method = read_method
+
+S3_CLIENT = boto3.client("s3")
+
+ROOT_CATALOG_ID = "root_catalog_id"
+ROOT_CATALOG_TITLE = "Geospatial Datalake Root Catalog"
+ROOT_CATALOG_DESCRIPTION = "The root catalog which links to all datasets catalogs in the data lake"
```
l0b0, line 16: The two values above should probably be changed to reflect the new product name, Geostore.

Should "datasets catalogs" be "dataset catalogs" (or even "catalogues")?

backend/write_to_catalog/task.py
```diff
@@ -0,0 +1,57 @@
...
+
+def lambda_handler(_event: JsonObject, _context: bytes) -> JsonObject:
+    """Main Lambda entry point."""
+
+    results = S3_CLIENT.list_objects(
+        Bucket=ResourceName.STORAGE_BUCKET_NAME.value, Prefix=CATALOG_KEY
+    )
+
+    # create root catalog if it doesn't exist
+    if "Contents" not in results:
```
l0b0, line 28: Since you have an `else` clause it would probably be easier to read this if you flip the clauses around to `if "Contents" in results:`. It basically avoids a double negative in the `else`.

Also, the "Contents" literal is used in two places in the code, here and in a test, so we should probably pull out a constant.

backend/write_to_catalog/task.py
```diff
@@ -0,0 +1,57 @@
...
+        root_catalog = Catalog(
+            id=ROOT_CATALOG_ID,
+            title=ROOT_CATALOG_TITLE,
+            description=ROOT_CATALOG_DESCRIPTION,
+            catalog_type=CatalogType.SELF_CONTAINED,
+        )
+
+    else:
+        root_catalog = Catalog.from_file(
+            f"s3://{ResourceName.STORAGE_BUCKET_NAME.value}/{CATALOG_KEY}"
```
l0b0, line 38: There's an `S3_URL_PREFIX` constant you can use for "s3://", although it will need to be pulled out into shared code.

Also, would it be possible to use the `result` above to get the file contents? I expect there should be something like a `results["Contents"][0]["Body"]` which is a file stream.

backend/write_to_catalog/task.py
```diff
@@ -0,0 +1,57 @@
...
+
+S3_CLIENT = boto3.client("s3")
+
+ROOT_CATALOG_ID = "root_catalog_id"
+ROOT_CATALOG_TITLE = "Geospatial Datalake Root Catalog"
+ROOT_CATALOG_DESCRIPTION = "The root catalog which links to all datasets catalogs in the data lake"
+CATALOG_KEY = "catalog.json"
+
+
+def lambda_handler(_event: JsonObject, _context: bytes) -> JsonObject:
```
l0b0, line 20: The underscore prefix in a parameter name is used to indicate (by convention only) that the argument value is unused, and is just there because whatever calls this code is out of our control. Since the event object is used it should probably not have an underscore prefix.

backend/write_to_catalog/task.py
```diff
@@ -0,0 +1,57 @@
...
+            description=ROOT_CATALOG_DESCRIPTION,
+            catalog_type=CatalogType.SELF_CONTAINED,
+        )
+
+    else:
+        root_catalog = Catalog.from_file(
+            f"s3://{ResourceName.STORAGE_BUCKET_NAME.value}/{CATALOG_KEY}"
+        )
+
+    for record in _event["Records"]:
```
l0b0, line 41: The "Records" literal is used in three places in the code, here and in tests, so we should probably pull out a constant.

backend/write_to_catalog/task.py
```diff
@@ -0,0 +1,57 @@
...
+        dataset_catalog = Catalog.from_file(
+            f"s3://{ResourceName.STORAGE_BUCKET_NAME.value}"
+            f"/{dataset_title_prefix}/{CATALOG_KEY}"
+        )
+
+        root_catalog.add_child(dataset_catalog)
+
+        root_catalog.normalize_hrefs(f"s3://{ResourceName.STORAGE_BUCKET_NAME.value}")
+        dataset_catalog.normalize_hrefs(
+            f"s3://{ResourceName.STORAGE_BUCKET_NAME.value}/{dataset_title_prefix}"
```
l0b0, line 52: Looks like this string could be pulled out and used in the `Catalog.from_file` call above.

infrastructure/constructs/processing.py
```diff
@@ -69,6 +83,33 @@ def __init__(
             "AmazonS3ReadOnlyAccess"
         )
 
+        ############################################################################################
+        # ROOT CATALOG UPDATE MESSAGE QUEUE
+
+        self.message_queue = aws_sqs.Queue(
+            self, "root-catalog-message-queue", visibility_timeout=Duration.seconds(60)
```
l0b0, line 90: Since this is tied to the Lambda timeout we should probably pull out a constant for it.

infrastructure/constructs/processing.py
```diff
@@ -69,6 +83,33 @@ def __init__(
...
+        )
+        self.message_queue_name_parameter = aws_ssm.StringParameter(
+            self,
+            "root catalog message queue name",
+            string_value=self.message_queue.queue_name,
+            description=f"Root Catalog Message Queue Name for {deploy_env}",
+            parameter_name=ParameterName.ROOT_CATALOG_MESSAGE_QUEUE_NAME.value,
+        )
+
+        write_catalog_lambda = BundledLambdaFunction(
```
l0b0, line 100: As discussed, do we maybe want to call this "add to" rather than "write"?

tests/aws_utils.py
```diff
@@ -276,6 +277,16 @@ class MockValidationResultFactory(Mock):
...
 
 
+def wait_for_s3_key(bucket_name: str, key: str, s3_client: S3Client) -> None:
+
+    process_timeout = datetime.now() + timedelta(minutes=3)
+    while (
+        "Contents" not in s3_client.list_objects(Bucket=bucket_name, Prefix=key)
+        and datetime.now() < process_timeout
+    ):
+        time.sleep(5)
```
l0b0, line 287: Should there be an `assert datetime.now() < process_timeout` instead, so that the test fails "early" (before trying to interact with the S3 file) if there is no such key?

tests/aws_utils.py
```diff
@@ -358,12 +369,14 @@ def wait_for_s3_batch_job_completion(
...
 ) -> DescribeJobResultTypeDef:
+    process_timeout = datetime.now() + timedelta(minutes=3)
+
     while (
         job_result := s3_control_client.describe_job(
             AccountId=account_id,
             JobId=s3_batch_job_arn,
         )
-    )["Job"]["Status"] not in S3_BATCH_JOB_FINAL_STATES:
+    )["Job"]["Status"] not in S3_BATCH_JOB_FINAL_STATES and datetime.now() < process_timeout:
```
l0b0, line 379: Should there be an `assert datetime.now() < process_timeout` instead, so that the test fails immediately if it never gets to this state?

tests/test_write_to_catalog.py
```diff
@@ -0,0 +1,139 @@
...
+
+        body = {"Records": [{"body": dataset.dataset_prefix}]}
+
+        try:
+            task.lambda_handler(body, any_lambda_context())
+
+            expected_links: JsonList = [
+                {
+                    STAC_REL_KEY: STAC_ROOT_KEY,
+                    STAC_HREF_KEY: f"./{CATALOG_KEY}",
```
l0b0, line 57: Is it possible to avoid including the "./" part at the start of these `href`s? I'm not sure it's valid in S3, and it's redundant.

tests/test_write_to_catalog.py
```diff
@@ -0,0 +1,139 @@
...
+        bucket_name=ResourceName.STORAGE_BUCKET_NAME.value,
+        key=f"{dataset.dataset_prefix}/{CATALOG_KEY}",
+    ):
+
+        body = {"Records": [{"body": dataset.dataset_prefix}]}
+
+        try:
+            task.lambda_handler(body, any_lambda_context())
+
+            expected_links: JsonList = [
```
l0b0, line 54: This variable can be created before running the `lambda_handler`, which makes it clearer that the data structure doesn't contain anything modified or returned by the `lambda_handler`.

tests/test_write_to_catalog.py
```diff
@@ -0,0 +1,139 @@
...
+            {
+                **deepcopy(MINIMAL_VALID_STAC_CATALOG_OBJECT),
+                STAC_ID_KEY: ROOT_CATALOG_ID,
+                STAC_DESCRIPTION_KEY: ROOT_CATALOG_DESCRIPTION,
+                STAC_TITLE_KEY: ROOT_CATALOG_TITLE,
+            }
+        ),
+        bucket_name=ResourceName.STORAGE_BUCKET_NAME.value,
+        key=CATALOG_KEY,
+    ):
```
l0b0, line 108: I think this test would be stronger if the original root catalogue already included some links, basically starting at the same state as the *end* state in the previous test.

l0b0 (changes requested):

backend/write_to_catalog/task.py
MitchellPaff: So far as I can see, the list_objects response does not contain file streams.

infrastructure/constructs/processing.py
```diff
@@ -1,4 +1,16 @@
-from aws_cdk import aws_dynamodb, aws_iam, aws_lambda_python, aws_s3, aws_ssm, aws_stepfunctions
+"""
+Data Lake processing stack.
```
l0b0, line 2: Is this a merge error?

tests/test_populate_catalog.py
```diff
@@ -0,0 +1,163 @@
...
+                    STAC_DESCRIPTION_KEY: ROOT_CATALOG_DESCRIPTION,
+                    STAC_TITLE_KEY: ROOT_CATALOG_TITLE,
+                    STAC_LINKS_KEY: expected_links,
+                }
+            ),
+            bucket_name=ResourceName.STORAGE_BUCKET_NAME.value,
+            key=CATALOG_KEY,
+        ):
+
+            expected_links.append(
```
l0b0, line 139: As discussed, rather than modifying this variable it would be good if it was instead assigned like `expected_links = original_links + [{…}]`. That way there's no confusion about the difference between the original and expected values.

l0b0 (changes requested):

l0b0 (approved):
Woot!

PR #741, opened 2021-05-25
feat: update dataset catalog with version

MitchellPaff:


backend/populate_catalog/task.py
```diff
@@ -18,11 +30,82 @@
...
+    return {}
+
+
+class UnhandledSQSMessageException(Exception):
+    pass
+
+
+class GeostoreSTACLayoutStrategy(HrefLayoutStrategy):
+    def get_catalog_href(self, cat: Catalog, parent_dir: str, is_root: bool) -> str:
+        original_path = cat.get_self_href().split("/")
```
l0b0, line 66: As discussed, we should probably use `urlparse(cat.get_self_href()).path` and `rsplit("/", maxsplit=2)` here.

backend/populate_catalog/task.py
```diff
@@ -18,11 +30,82 @@
...
+
+    def get_item_href(self, item: Item, parent_dir: str) -> str:
+        original_path = item.get_self_href().split("/")
+        return os.path.join(parent_dir, original_path[-1])
+
+
+def handle_dataset(version_metadata_key: str) -> None:
+    """Handle writing a new dataset version to the dataset catalog"""
+    storage_bucket_path = f"{S3_URL_PREFIX}{ResourceName.STORAGE_BUCKET_NAME.value}"
+    dataset_prefix = version_metadata_key.split("/", maxsplit=2)[0]
```
l0b0, line 91: `maxsplit=2` gives you up to *three* parts. Since we only care about the first one I think we can use `maxsplit=1` here.

backend/populate_catalog/task.py
```diff
@@ -1,10 +1,22 @@
+import os
```
l0b0, line 1: You can shorten the code below a bit by using `from os.path import join`.

backend/sqs_message_attributes.py
```diff
@@ -0,0 +1,11 @@
+def decapitalize(key: str) -> str:
+    return f"{key[:1].lower()}{key[1:]}"
```
l0b0, line 2: Maybe add a comment about why we need this?

tests/test_populate_catalog.py
```diff
@@ -136,28 +169,384 @@ def should_update_existing_root_catalog(subtests: SubTests) -> None:
...
                     STAC_TYPE_KEY: "application/json",
                 }
             ]
 
-            body = {RECORDS_KEY: [{"body": dataset.dataset_prefix}]}
+            expected_dataset_links: JsonList = [
+                {
+                    STAC_REL_KEY: STAC_REL_ROOT,
+                    STAC_HREF_KEY: f"../{CATALOG_KEY}",
+                    STAC_TYPE_KEY: "application/json",
```
l0b0, line 184: There doesn't seem to be any constants for MIME types; should we pull one out for this extremely common one?

l0b0 (changes requested):

backend/populate_catalog/task.py
```diff
@@ -18,11 +31,79 @@
...
+            cat_root = parent_dir
+        else:
+            cat_root = join(parent_dir, original_path[-2])
+
+        return join(cat_root, original_path[-1])
+
+    def get_collection_href(self, col: Collection, parent_dir: str, is_root: bool) -> str:
+        original_path = urlparse(col.get_self_href()).path.rsplit("/", maxsplit=2)
+        assert not is_root
+        col_root = join(parent_dir, original_path[-2])
```
l0b0, line 78: You can join more than two file paths, so you could `return join(col_root, *original_path[-2:])`

backend/stac_format.py
```diff
@@ -10,16 +10,20 @@
 STAC_HREF_KEY = "href"
 STAC_ID_KEY = "id"
 STAC_LICENSE_KEY = "license"
+STAC_LINKS_KEY = "links"
+STAC_MEDIA_TYPE_GEOJSON = "application/geo+json"
```
l0b0, line 14: This doesn't seem to be used anywhere.

l0b0 (approved):

PR #752, opened 2021-05-27
refactor: Pass dataset prefix earlier in pipeline to avoid fetching in multiple places

MitchellPaff:
Refactored the input to pipeline to include the dataset prefix.
This is to avoid fetching this from the database multiple times during
the pipeline.


backend/check_stac_metadata/task.py
```diff
@@ -36,10 +36,11 @@ def lambda_handler(event: JsonObject, _context: bytes) -> JsonObject:
                 "type": "object",
                 "properties": {
                     DATASET_ID_KEY: {"type": "string"},
+                    DATASET_PREFIX_KEY: {"type": "string"},
```
l0b0, line 39: As discussed, this property isn't used by this task, so it shouldn't be in the schema. Ditto in other tasks.

l0b0 (changes requested):

l0b0 (approved):

PR #802, opened 2021-06-16
refactor: Remove cdk bootstrapping command from pipelines

MitchellPaff:
Removing from CI/CD to a cron job that will regularly bootstrap our AWS accounts to the latest versions


.github/workflows/cdk-bootstrap.yml
```diff
@@ -0,0 +1,92 @@
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ap-southeast-2
+          mask-aws-account-id: true
+          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_PROD }}
+
+      - name: Print CDK version
+        run: poetry run cdk --version
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
l0b0, line 88: We shouldn't need to print the CDK version or check out the repo more than once.

.github/workflows/cdk-bootstrap.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Bootstrap CDK
+        run: |
+          poetry run cdk bootstrap aws://unknown-account/ap-southeast-2
+
+      - name: (Prod) Configure AWS credentials
```
l0b0, line 75: Should we maybe make each of the bootstraps a separate job? Then they can run in parallel, and should all end up in the same state rather than a failure in one step resulting in the subsequent environments not being updated at all.

.github/workflows/cdk-bootstrap.yml
```diff
@@ -0,0 +1,150 @@
...
+        run: npm ci --production
+
+      - name: Add local Node packages to PATH
+        run: echo "./node_modules/.bin:$PATH" >> $GITHUB_PATH
+
+      - name: Print CDK version
+        run: poetry run cdk --version
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
l0b0, line 137: This should probably be merged with the other `actions/checkout@v2`.

l0b0 (approved):

PR #825, opened 2021-06-23
feat: Add functionality to notify slack when Step Function completes

MitchellPaff:
Still need to write documentation


infrastructure/constructs/notify.py
```diff
@@ -0,0 +1,73 @@
...
+            handler="notify_slack.task.lambda_handler",
+            runtime=PYTHON_RUNTIME,
+            timeout=LAMBDA_TIMEOUT,
+            code=bundled_code("notify_slack"),
+            layers=[botocore_lambda_layer],  # type: ignore[list-item]
+        )
+        if SLACK_NOTIFY_URL_ENV_NAME in environ:
+            slack_notify_function.add_environment(
+                SLACK_NOTIFY_URL_ENV_NAME,
+                environ.get(SLACK_NOTIFY_URL_ENV_NAME),  # type: ignore[arg-type]
```
l0b0, line 44: You know the environment variable exists here, so you can use `environ[SLACK_NOTIFY_URL_ENV_NAME]`. This also makes the value no longer `Optional`, so you can probably remove the annotation as well.

infrastructure/constructs/notify.py
```diff
@@ -0,0 +1,73 @@
...
+        scope: Construct,
+        stack_id: str,
+        *,
+        botocore_lambda_layer: aws_lambda_python.PythonLayerVersion,
+        env_name: str,
+        state_machine_arn: str,
+    ) -> None:
+        super().__init__(scope, stack_id)
+
+        slack_notify_function = aws_lambda.Function(
```
l0b0, line 31: Would it be possible to use `BundledLambdaFunction` here?

backend/notify_slack/task.py
```diff
@@ -0,0 +1,38 @@
...
+
+def lambda_handler(event: JsonObject, _context: bytes) -> JsonObject:
+    LOGGER.debug(dumps({EVENT_KEY: event}))
+    print(event)
+    response_body = get_import_status_given_arn(event[STEP_FUNCTION_ARN_KEY])
+
+    if SLACK_NOTIFY_URL_ENV_NAME in environ:
+        http.request(
+            "POST",
+            environ.get(SLACK_NOTIFY_URL_ENV_NAME),
```
l0b0, line 30: [Ditto](https://github.com/linz/geostore/pull/825#discussion_r657471376)

backend/import_status/get.py
```diff
@@ -1,39 +1,22 @@
...
-else:
-    # In production we want to avoid depending on a package which has no runtime impact
-    SFNClient = object
-
-STEP_FUNCTIONS_CLIENT: SFNClient = boto3.client("stepfunctions")
 LOGGER = set_up_logging(__name__)
 
 
 def get_import_status(body: JsonObject) -> JsonObject:
-    LOGGER.debug(dumps({"event": body}))
```
l0b0, line 36: Good catch, but does this maybe belong in a separate PR? There are a bunch more occurrences of this, according to `git grep '"event"'`.

backend/notify_slack/task.py
```diff
@@ -0,0 +1,44 @@
...
+import urllib3  # type: ignore[import]
+
+from ..api_keys import EVENT_KEY
+from ..log import set_up_logging
+from ..step_function import get_import_status_given_arn
+from ..types import JsonObject
+
+http = urllib3.PoolManager()
+
+SLACK_CHANNEL_NAME = "#geostore-notifications"
```
l0b0, line 13: Should this perhaps be an environment variable, configured in cdk.json?

tests/test_notify_slack.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+@patch("backend.notify_slack.task.http.request")
+@patch("backend.notify_slack.task.get_import_status_given_arn")
+def should_notify_slack_when_url_set(
+    get_import_status_mock: MagicMock, http_mock: MagicMock, subtests: SubTests
+) -> None:
+    # Given
+
+    get_import_status_mock.return_value = get_import_status_output = {
```
l0b0, line 44: I don't think the contents of the `get_import_status_given_arn` response is relevant to these tests, because `lambda_handler` simply returns the response verbatim. So I think you could just do `get_import_status_mock.return_value = get_import_status_output = {}`.

tests/test_notify_slack.py
```diff
@@ -0,0 +1,129 @@
...
+    METADATA_UPLOAD_KEY,
+    STATUS_KEY,
+    STEP_FUNCTION_KEY,
+    VALIDATION_KEY,
+)
+
+from .aws_utils import any_arn_formatted_string, any_lambda_context
+from .general_generators import any_https_url
+
+LOGGER = getLogger("backend.notify_slack.task")
```
l0b0, line 34: This is only used in one test, so it might as well be part of it.

l0b0 (changes requested):

backend/import_status/get.py
MitchellPaff: will follow up with another PR

l0b0 (dismissed):

l0b0 (approved):

PR #837, opened 2021-06-28
feat: add access for OpenTopography.org

imincik:
This change adds S3 bucket read role for OpenTopography access.


README.md
```diff
@@ -40,6 +40,7 @@ settings in cdk.json. To disable these, simply remove the context entries or set
 The settings are:
 
 - `enableLDSAccess`: if true, gives the LINZ Data Service/Koordinates access to the storage bucket.
+- `enableOpenTopographyAccess`: if true, gives the OpenTopography access to the storage bucket.
```
l0b0, line 43: "the" is superfluous here. Also, both of these settings should specify that they give *read-only* access.

infrastructure/constructs/opentopo.py
```diff
@@ -0,0 +1,25 @@
...
+from .roles import MAX_SESSION_DURATION
+
+
+class OpenTopography(Construct):
+    def __init__(
+        self, scope: Construct, stack_id: str, *, env_name: str, storage_bucket: aws_s3.Bucket
+    ) -> None:
+        super().__init__(scope, stack_id)
+
+        account_principal = aws_iam.AccountPrincipal(account_id=99)  # TODO: add OT AWS account id
```
l0b0, line 13: We probably need to wait until we have the account ID before merging. Don't want to accidentally give some random account access, or fail because that ID doesn't exist.

l0b0 (changes requested):

README.md
imincik: Fixed now. Thanks

infrastructure/constructs/opentopo.py
imincik: AWS account updated now.

l0b0 (approved):

PR #859, opened 2021-07-02
fix: handle step function returning null on failure

MitchellPaff:
Didn't realise that the step function "output" is null when it fails - have to pull or generate data from elsewhere to allow failures to still show validation results etc.

### DIFF
--- a/tests/test_notify_slack.py
+++ b/tests/test_notify_slack.py
@@ -20,20 +20,25 @@ from backend.notify_status_update.task import (
     STEP_FUNCTION_OUTPUT_KEY,
     STEP_FUNCTION_STARTDATE_KEY,
     STEP_FUNCTION_STOPDATE_KEY,
-    STEP_FUNCTION_UPDATE_DATASET_KEY,
-    STEP_FUNCTION_UPLOAD_STATUS_KEY,
     WEBHOOK_MESSAGE_BLOCKS_KEY,
     lambda_handler,
     publish_sns_message,
 )
+from backend.step_function import Outcome
 from backend.step_function_keys import (
     ASSET_UPLOAD_KEY,
     DATASET_ID_KEY,
     DATASET_PREFIX_KEY,
+    ERRORS_KEY,
+    JOB_STATUS_FAILED,
+    JOB_STATUS_RUNNING,
     JOB_STATUS_SUCCEEDED,
     METADATA_UPLOAD_KEY,
     NEW_VERSION_S3_LOCATION,
     STATUS_KEY,
+    STEP_FUNCTION_KEY,
+    UPDATE_DATASET_KEY,
+    UPLOAD_STATUS_KEY,
     VALIDATION_KEY,
     VERSION_ID_KEY,
 )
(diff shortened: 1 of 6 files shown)


tests/test_notify_slack.py
```diff
@@ -113,6 +131,54 @@ def should_not_notify_slack_when_step_function_running(webhook_client_mock: Magi
         webhook_client_mock.assert_not_called()
 
 
+@patch("backend.notify_status_update.task.WebhookClient.send")
+@patch("backend.notify_status_update.task.get_import_status_given_arn")
+def should_not_notify_slack_when_step_function_failed(
```
l0b0, line 136: Is this title accurate? The assertion is that `webhook_client_mock` *was* called. If it is accurate, why would we not want to notify on failure?

tests/test_notify_slack.py
```diff
@@ -113,6 +131,54 @@ def should_not_notify_slack_when_step_function_running(webhook_client_mock: Magi
...
+    step_func_status_mock: MagicMock, webhook_client_mock: MagicMock
+) -> None:
+    # Given
+
+    webhook_client_mock.return_value.status_code = HTTPStatus.OK
+    mock_slack_url = any_https_url()
+    now_ts = datetime.now()
```
l0b0, line 143: Using the actual current datetime in tests is often a problem, in particular because it means the test is guaranteed not reproducible. Would it be possible to use a static datetime here, such as `datetime(2001, 2, 3, hour=4, minute=5, second=6, microsecond=789876, tzinfo=timezone.utc)` used elsewhere? Same in `should_notify_slack_with_finished_details_when_url_set`.

tests/test_notify_slack.py
```diff
@@ -113,6 +131,54 @@ def should_not_notify_slack_when_step_function_running(webhook_client_mock: Magi
...
+                STEP_FUNCTION_STOPDATE_KEY: round(now_ts.timestamp() * 1000),
+            },
+            STEP_FUNCTION_OUTPUT_KEY: None,
+        }
+
+        lambda_handler(notify_status_update_input, any_lambda_context())
+
+        # Then
+        webhook_client_mock.assert_called_once()
+        assert len(webhook_client_mock.call_args[1][WEBHOOK_MESSAGE_BLOCKS_KEY]) == 9
```
l0b0, line 179: I don't understand this assertion (and the corresponding one in `should_notify_slack_with_finished_details_when_url_set`). Would it be easier to understand if we asserted the actual values of (a subset of) the arguments? If there's no easy way to make this readable to someone unfamiliar with `WebhookClient.send`, could you please add some context to the "Then" comment above?

l0b0 (changes requested):

tests/test_notify_slack.py
MitchellPaff: Hey, yeah It's asserting the number of message 'blocks' (concept from slack_sdk) sent in the request. I decided to do this cause it would potentially be overkill to assert the blocks individually. I could assert some text values from the blocks, for example that the first block text is "Geostore Dataset Version Import" etc. What do you think?

tests/test_notify_slack.py
```diff
@@ -42,18 +47,41 @@
 from .general_generators import any_https_url
 from .stac_generators import any_dataset_id, any_dataset_prefix, any_dataset_version_id
 
+STEP_FUNCTION_START_DATE = round(
+    datetime(
+        2001, 2, 3, hour=4, minute=5, second=6, microsecond=789876, tzinfo=timezone.utc
+    ).timestamp()
+    * 1000
+)
+STEP_FUNCTION_STOPDATE = round(
```
l0b0, line 56: You could just do `STEP_FUNCTION_STOPDATE = STEP_FUNCTION_START_DATE + some_number` here.

tests/test_notify_slack.py
```diff
@@ -42,18 +47,41 @@
 from .general_generators import any_https_url
 from .stac_generators import any_dataset_id, any_dataset_prefix, any_dataset_version_id
 
+STEP_FUNCTION_START_DATE = round(
```
l0b0, line 50: Rather than "date" this should probably be "millisecond timestamp".

l0b0 (changes requested):

tests/test_notify_slack.py
l0b0: Ah, right. Nah, it's a bit clearer now.

l0b0 (approved):

PR #868, opened 2021-07-06
fix: add infrastructure test to notify function

MitchellPaff:


tests/test_notify_slack.py
```diff
@@ -238,3 +243,37 @@ def should_publish_sns_message(get_param_mock: MagicMock) -> None:
...
+@mark.infrastructure
+def should_launch_notify_slack_endpoint_lambda_function(
+    lambda_client: LambdaClient, events_client: EventBridgeClient
+) -> None:
+
+    cloudwatch_events_response = events_client.list_targets_by_rule(
+        Rule=ResourceName.CLOUDWATCH_RULE_NAME.value
+    )
+
+    notify_status_lambda_arn = cloudwatch_events_response["Targets"][0]["Arn"]
```
l0b0, line 257: Are these keys constants we could pull out? Ditto for `"Payload"` below.

l0b0 (approved):

PR #879, opened 2021-07-09
feat: Add user documentation for subscribing to an SNS topic

MitchellPaff:


USAGE.md
```diff
@@ -246,3 +246,80 @@ $ aws lambda invoke --function-name import-status \
 
 {"status_code": 200, "body": {"validation":{ "status": "SUCCEEDED"}, "metadata_upload":{"status": "Pending", "errors":[]}, "asset_upload":{"status": "Pending", "errors":[]}}}
 ```
+
+## Receive Import Status updates by subscribing to our AWS SNS Topic
+
+The ARN of our SNS Topic is `arn:aws:sns:ap-southeast-2:715898075157:geostore-import-status` which
+you may choose to subscribe too.
```
l0b0: "too" should be "to".

USAGE.md
```diff
@@ -246,3 +246,80 @@ $ aws lambda invoke --function-name import-status \
...
+
+You may also choose to apply a subscription filter policy, which will filter notifications for a
+specific dataset or specific statuses.
+
+```json
+{
+  "dataset_id": ["id_1234567890", "id_0987654321"],
```
l0b0, line 260: It might be useful to include realistic dataset IDs here for clarity.

USAGE.md
```diff
@@ -246,3 +246,80 @@ $ aws lambda invoke --function-name import-status \
...
+  "status": ["RUNNING", "SUCCEEDED", "FAILED", "TIMED_OUT", "ABORTED"]
```
l0b0, line 261: Are these all the possible states? If so, you might want to specify that above.

USAGE.md
```diff
@@ -246,3 +246,80 @@ $ aws lambda invoke --function-name import-status \
...
+  "Message": "MESSAGE BELOW",
+  "Timestamp": "2021-07-07T01:49:33.471Z",
+  "SignatureVersion": "1",
+  "Signature": "xxxxx",
+  "SigningCertURL": "https://example.com",
+  "UnsubscribeURL": "https://example.com",
+  "MessageAttributes": {
+    "dataset_id": {
+      "Type": "String",
+      "Value": "test_123456789"
```
l0b0, line 282: Does this correspond to the dataset IDs in the filter above?

USAGE.md
```diff
@@ -246,3 +246,80 @@ $ aws lambda invoke --function-name import-status \
...
+    "status": {
+      "Type": "String",
+      "Value": "SUCCEEDED"
+    }
+  }
+}
+```
+
+Below is an example of the EventBridge 'Step Functions Execution Status Change' payload. This will
+be formatted as a string and associated with the 'Message' attribute in the above SNS payload. See
```
l0b0: How is it "associated with the 'Message' attribute in the above SNS payload"? Do you mean that below is a JSON representation of the string shown as "MESSAGE BELOW" above? If so, maybe something like `{"version": "0", […]}` in the above would be useful to show the link more directly?

USAGE.md
```diff
@@ -246,3 +246,80 @@ $ aws lambda invoke --function-name import-status \
...
+  "version": "0",
+  "id": "3afc69d1-1291-6372-ae76-9a9c11ee35c4",
+  "detail-type": "Step Functions Execution Status Change",
+  "source": "aws.states",
+  "account": "715898075157",
+  "time": "2021-07-07T01:49:29Z",
+  "region": "ap-southeast-2",
+  "resources": ["arn:aws:states:example"],
+  "detail": {
+    "executionArn": "arn:aws:states:example",
```
l0b0, line 308: Is this ARN actually in the "states" namespace? If not it might be better to ellipsise it (`[…]`) or use the real namespace.

USAGE.md
```diff
@@ -246,3 +246,80 @@ $ aws lambda invoke --function-name import-status \
...
+    },
+    "output": "{\"dataset_id\":\"01F9ZA9ZZZDM815S20EHXEAT40\",\"dataset_prefix\":\"test_1625622377-01F9ZA9ZZZDM815S20EHXEAT40\",\"version_id\":\"2021-07-07T01-46-30-787Z_9NJEAD3VXRCH5W05\",\"metadata_url\":\"s3://example/catalog.json\",\"s3_role_arn\":\"arn:aws:iam::715898075157:role/example\",\"content\":{\"first_item\":\"0\",\"iteration_size\":1,\"next_item\":-1,\"assets_table_name\":\"example\",\"results_table_name\":\"example\"},\"validation\":{\"success\":true},\"import_dataset\":{\"asset_job_id\":\"e4ad8b0d-4358-4c42-bb0d-3577c96f7039\",\"metadata_job_id\":\"84a7b4fc-7d00-403c-a5fb-91257f406afb\"},\"upload_status\":{\"validation\":{\"status\":\"Passed\",\"errors\":[]},\"asset_upload\":{\"status\":\"Complete\",\"errors\":[]},\"metadata_upload\":{\"status\":\"Complete\",\"errors\":[]}},\"update_dataset_catalog\":{\"new_version_s3_location\":\"s3://linz-geostore/example/catalog.json\"}}",
+    "outputDetails": {
+      "included": true
+    }
+  }
+}
+```
+
+Note: the output field will only be populated above when the Step Function has succeeded.
```
l0b0, line 326: Should "succeeded" be "finished" in this line?

l0b0 (changes requested):

USAGE.md
MitchellPaff: yes.

USAGE.md
MitchellPaff: yes it is.

USAGE.md
MitchellPaff: No, if the step function fails it won't have an output.

l0b0 (approved):

PR #882, opened 2021-07-12
docs: Add further user documentation for SNS message filtering

MitchellPaff:

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -249,15 +249,19 @@ $ aws lambda invoke --function-name import-status \
 
 ## Receive Import Status updates by subscribing to our AWS SNS Topic
 
-The ARN of our SNS Topic is `arn:aws:sns:ap-southeast-2:715898075157:geostore-import-status` which
-you may choose to subscribe to.
+The ARN of our SNS Topic is `arn:aws:sns:ap-southeast-2:<YOUR_ACCOUNT_ID>:geostore-import-status`
+which you may choose to subscribe to.
 
 You may also choose to apply a subscription filter policy, which will filter notifications for a
 specific dataset or specific statuses. Included in the example is all the valid statuses.
 
+The Geostore will store a dataset in a top level directory name with format of (dataset
+title)-(dataset id) You can filter SNS topics for a dataset by providing the entire dataset
+directory name or by providing just the title in a 'prefix' object. Examples below.
+
 ```json
 {
-  "dataset_id": ["01F9ZGZ4EGEZ9R4Z48VNQHKB0Z", "01F9ZFRK12V0WFXJ94S0DHCP65"],
+  "dataset_id": [{ "prefix": "Taranaki" }, "Taranaki_2020-01F9ZFRK12V0WFXJ94S0DHCP65"],
   "status": ["RUNNING", "SUCCEEDED", "FAILED", "TIMED_OUT", "ABORTED"]
 }
 ```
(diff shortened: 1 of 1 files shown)


USAGE.md
```diff
@@ -249,15 +249,19 @@ $ aws lambda invoke --function-name import-status \
...
 specific dataset or specific statuses. Included in the example is all the valid statuses.
 
+The Geostore will store a dataset in a top level directory name with format of (dataset
+title)-(dataset id) You can filter SNS topics for a dataset by providing the entire dataset
+directory name or by providing just the title in a 'prefix' object. Examples below.
+
 ```json
 {
-  "dataset_id": ["01F9ZGZ4EGEZ9R4Z48VNQHKB0Z", "01F9ZFRK12V0WFXJ94S0DHCP65"],
+  "dataset_id": [{ "prefix": "Taranaki" }, "Taranaki_2020-01F9ZFRK12V0WFXJ94S0DHCP65"],
```
l0b0, line 264: What's going on here? Are we actually returning a `List[Union[Dict[str, str], str]]`?

USAGE.md
```diff
@@ -249,15 +249,19 @@ $ aws lambda invoke --function-name import-status \
 
 ## Receive Import Status updates by subscribing to our AWS SNS Topic
 
-The ARN of our SNS Topic is `arn:aws:sns:ap-southeast-2:715898075157:geostore-import-status` which
-you may choose to subscribe to.
+The ARN of our SNS Topic is `arn:aws:sns:ap-southeast-2:<YOUR_ACCOUNT_ID>:geostore-import-status`
```
billgeo, line 252: Isn't this supposed to say <GEOSTORE_AWS_ACCOUNT_ID>?

USAGE.md
```diff
@@ -300,7 +304,7 @@ See
   "id": "3afc69d1-1291-6372-ae76-9a9c11ee35c4",
   "detail-type": "Step Functions Execution Status Change",
   "source": "aws.states",
-  "account": "715898075157",
+  "account": "<YOUR_ACCOUNT_ID>",
```
billgeo, line 307: Same here

USAGE.md
```diff
@@ -269,7 +273,7 @@ contain a JSON string with specific details regarding the Step Function Executio
 {
   "Type": "Notification",
   "MessageId": "xxxx-xxxx-xxxx",
-  "TopicArn": "arn:aws:sns:ap-southeast-2:715898075157:geostore-import-status",
+  "TopicArn": "arn:aws:sns:ap-southeast-2:<YOUR_ACCOUNT_ID>:geostore-import-status",
```
billgeo, line 276: Same here.

USAGE.md
billgeo: Also a little confused. I would have thought it would be something like
`"dataset_id": [{ "prefix": "Taranaki" }]`

billgeo (changes requested):

USAGE.md
MitchellPaff: We're  not returning anything. This is an example json policy that can be used to filter the SNS messages - it has one example where a prefix is used one where the whole string is matched.

billgeo (approved):
LGTM. Thanks.

billgeo (changes requested):
Actually the account ID also appears further down in the  example.
Can you make sure there sensitive info anywhere.

billgeo (approved):
Sorry, ignore that last comment. You changed it.

PR #901, opened 2021-07-14
fix: correctly report s3 copy job failures

MitchellPaff:


tests/test_import_status_endpoint.py
```diff
@@ -168,10 +175,7 @@ def should_report_s3_batch_upload_failures(
     }
 
     describe_s3_job_mock.return_value = {
-        "Job": {
-            "Status": "Completed",
-            "FailureReasons": [{"FailureCode": "TEST_CODE", "FailureReason": "TEST_REASON"}],
-        }
+        "Job": {"Status": S3_BATCH_STATUS_COMPLETE, "ProgressSummary": {"NumberOfTasksFailed": 1}}
```
l0b0, line 178: Shouldn't we still keep the original test with `FailureReasons`? I feel like `NumberOfTasksFailed` warrants a separate but very similar test.

tests/test_import_status_endpoint.py
```diff
@@ -144,6 +152,68 @@ def should_retrieve_validation_failures(describe_step_function_mock: MagicMock)
...
+        BODY_KEY: {
+            STEP_FUNCTION_KEY: {STATUS_KEY: "Succeeded"},
+            VALIDATION_KEY: {STATUS_KEY: Outcome.PASSED.value, ERRORS_KEY: []},
+            METADATA_UPLOAD_KEY: {
+                STATUS_KEY: S3_BATCH_STATUS_FAILED,
+                ERRORS_KEY: {FAILED_TASKS_KEY: 1, FAILURE_REASONS_KEY: []},
+            },
+            ASSET_UPLOAD_KEY: {
+                STATUS_KEY: S3_BATCH_STATUS_FAILED,
+                ERRORS_KEY: {FAILED_TASKS_KEY: 1, FAILURE_REASONS_KEY: []},
```
l0b0, line 197: We should make sure the failed tasks are reported independently for assets and metadata files. You should be able to do this by creating a function which mimics [`describe_job`](https://boto3.amazonaws.com/v1/documentation/api/1.17.61/reference/services/s3control.html#S3Control.Client.describe_job) and returns two different values, something like this:

```python
def describe_job_mock(**kwargs: Any):
    if kwargs["JobId"] == metadata_job_id:
        return metadata_failed_task_count
    if kwargs["JobId"] == asset_job_id:
        return asset_failed_task_count
    assert False, f"Unknown job ID {kwargs['JobId']}"

describe_s3_job_mock.side_effect = describe_job_mock
```

l0b0 (changes requested):

l0b0 (approved):

PR #922, opened 2021-07-27
feat: Support Nix

l0b0:
Feature checklist:

- [x] Build verified in pipeline
- [x] Documentation
- [x] Direnv loads Nix shell when `cd`-ing into the directory
- [x] Auto-formatting
- [x] `cdk synth` runs
- [x] `pytest` runs
- [x] `coverage run --module pytest` runs
- [x] `poetry lock` runs
- [x] Locks package repo version
- [x] Locks package repo to a *stable* version
- [ ] Build Node.js packages with Nix (postponed)
- [ ] Node.js package executables on `$PATH` immediately after entering new shell from scratch (postponed)

## Issues

Closes https://github.com/linz/geostore/issues/996

## Reference


imincik (commented):
As I said in Slack #x-nix, I am not 100% convinced to hop on `poetry2nix` but I very much like the idea of using `nix-shell`. That's why I am thinking about manually using Poetry inside of `nix-shell` until `poetry2nix` will be more mature and will contain good support for all packages we need without need of any overrides.

imincik (commented):
Also, I am not sure if running `npm install` is a good idea in shell hook. Isn't it going to delay every nix shell entrance ?

l0b0:
> Also, I am not sure if running `npm install` is a good idea in shell hook. Isn't it going to delay every nix shell entrance ?

It will, but I'm not familiar enough with Nix to know which solution to use for building Node packages as part of the shell.

imincik:
When running `nix-shell` I saw it is downloading two versions of Python

```
copying path '/nix/store/66p7i56lg3zzp7nj6g0fywb704ihp4iy-python3-3.8.11' from 'https://cache.nixos.org'...
copying path '/nix/store/1g54cwl6m78rgrqsnn1pp29dbvldpvsd-python3-3.9.6' from 'https://cache.nixos.org'...
```

and many of packages are downloaded in two versions as well

```
copying path '/nix/store/hw9964pszkfqj424sbiawkz6i4h311l2-six-1.15.0.tar.gz' from 'https://cache.nixos.org'...
copying path '/nix/store/qzfsn8cfr9pfk586iiq5hrcbhcnmv236-six-1.16.0.tar.gz' from 'https://cache.nixos.org'...
copying path '/nix/store/grpnvb8w4n0n1fd7lrw5rnff67wp9f8h-source' from 'https://cache.nixos.org'...
copying path '/nix/store/mvpfgj392km5qvzvhx1slk973nvj9bqp-source' from 'https://cache.nixos.org'...
copying path '/nix/store/28ba35p0qv3lrjgpxd6dz4kfjfryr55l-stdenv-linux' from 'https://cache.nixos.org'...
copying path '/nix/store/r4bl79l2bdjawmr2rhhqvci56qh0fkvv-stdenv-linux' from 'https://cache.nixos.org'...
```

Do you have any idea why is it ?

imincik:
`nix-shell` command finished with following in my case:

```
warning: dumping very large path (> 256 MiB); this may run out of memory
Found '/home/imincik/Projects/dev/linz/data-lake3/code/.nvmrc' with version <14.17.0>
N/A: version "14.17.0 -> N/A" is not yet installed.

You need to run "nvm install 14.17.0" to install it before using it.
```

and it didn't enter nix-shell

l0b0:
> `nix-shell` command finished with following in my case:
> 
> ```
> warning: dumping very large path (> 256 MiB); this may run out of memory
> Found '/home/imincik/Projects/dev/linz/data-lake3/code/.nvmrc' with version <14.17.0>
> N/A: version "14.17.0 -> N/A" is not yet installed.
> 
> You need to run "nvm install 14.17.0" to install it before using it.
> ```
> 
> and it didn't enter nix-shell

That must be `nvm` interfering in some way - I don't have it installed. Could you try disabling it?

imincik:
> That must be nvm interfering in some way.

Yes, it was because of my `nvm` installation.

imincik:
For me the biggest con of using `poetry2nix` is that it really takes a very long time to build it. Looks like at least 30 minutes on my machine.

l0b0:
> For me the biggest con of using `poetry2nix` is that it really takes a very long time to build it. Looks like at least 30 minutes on my machine and the output is not very nice (as is Poetry has)

Yes, that's definitely a drawback. If we end up settling on Nix long term, though, we can get rid of Poetry completely and just configure the dependencies in Nix as well.

imincik:
> > That must be nvm interfering in some way.
> 
> I still have that `nvm` issue. Just trying to find where the problem is

I believe I have `nvm` installed according instructions we have in README

imincik:
The problem is that `nvm` is used in `activate-dev-env.bash`

imincik:
Once I manually installed Node using `nvm install` I stopped getting this error but I guess this is not how we want to build Nix env. Node should be picked up from Nix

l0b0:
> Once I manually installed Node using `nvm install` I stopped getting this error but I guess this is not how we want to build Nix env. Node should be picked up from Nix

If you have other software which overrides how Nix works then it's not going to work. But we should probably use Nix to install the Node packages directly rather than installing Node.js and running `npm install` manually.

shell.nix
```diff
@@ -0,0 +1,26 @@
...
+    python = pkgs.python38;
+    projectDir = builtins.path { path = ./.; name = "geostore"; };
+  };
+in
+pkgs.mkShell {
+  buildInputs = [
+    pkgs.nodejs-14_x
+    poetryEnv
+    pkgs.python38Packages.pip
+    pkgs.python38Packages.poetry
```
imincik, line 21: `pkgs.poetry` should be used here instead of `pkgs.python38Packages.poetry`. `pythonXYPackages.xyz` packages should be used when building Python libraries closure. What you want here is just `poetry` exacutable

shell.nix
```diff
@@ -0,0 +1,26 @@
...
+  poetryEnv = pkgs.poetry2nix.mkPoetryEnv {
+    python = pkgs.python38;
+    projectDir = builtins.path { path = ./.; name = "geostore"; };
+  };
+in
+pkgs.mkShell {
+  buildInputs = [
+    pkgs.nodejs-14_x
+    poetryEnv
+    pkgs.python38Packages.pip
```
imincik, line 20: I am not sure if you need `pip` here.

l0b0: I did add it for a reason, but I can't remember now. I've removed it; we can always add it back in if necessary.

imincik: Was is because of this error ?

```

  ModuleNotFoundError

  No module named 'virtualenv.activation.nushell'

  at <frozen importlib._bootstrap>:973 in _find_and_load_unlocked
```

l0b0: No, where do you see that?

PR #989, opened 2021-09-21
feat: Validate LINZ extension

l0b0:
## Issues

Closes https://github.com/linz/geostore/issues/78.

## Reference


.gitmodules
```diff
@@ -2,6 +2,9 @@
 	path = backend/check_stac_metadata/geojson-spec
 	url = https://github.com/geojson/schema.git
 	branch = gh-pages
+[submodule "stac"]
```
l0b0, line 5: FYI, this name is maybe a bit weird. I originally called it `linz-stac`, but using the original name meant that I could use `LATEST_LINZ_SCHEMA_DIRECTORY` all over the place.

backend/stac_format.py
```diff
@@ -26,4 +28,12 @@
 STAC_TYPE_COLLECTION = "Collection"
 STAC_TYPE_ITEM = "Feature"
 STAC_TYPE_KEY = "type"
+STAC_UPDATED_KEY = "updated"
+STAC_VERSION = "1.0.0"
 STAC_VERSION_KEY = "stac_version"
+
+LATEST_LINZ_STAC_VERSION = "0.0.3"
```
l0b0, line 35: We could get this by version ordering the directories in the submodule. But it would be a fairly expensive operation, and hopefully Dependabot updating the submodule should trigger `should_treat_linz_example_json_files_as_valid` to fail if there are any semantic changes to the repo.

l0b0: @billgeo Are we definitely going to need to support more than one version of the LINZ STAC extension?

billgeo: In the near future, I can't think of a case where we'd need to support older versions. Seems like this is all about compliance and ensuring all metadata is consistent with the latest version.

MitchellPaff (commented):
I had to make some changes to field names on https://github.com/linz/stac/pull/45

mainly prefiing the created and updated fields with "linz:".

Not sure if this directly affects this PR?

MitchellPaff (approved):

PR #990, opened 2021-09-21
refactor: Run non-infrastructure tests before configuring AWS

l0b0:
This way we can make sure the correct tests are marked as
"infrastructure".

## Reference


l0b0:
Annoyingly, since some of the AWS code runs at import time the fix for this might be a fair bit more involved.

PR #1173, opened 2021-11-09
fix: correct logging format

MitchellPaff:
## Reference


geostore/logging_keys.py
```diff
@@ -0,0 +1,6 @@
+LOG_MESSAGE_LAMBDA_START = "Lambda:Start"
+LOG_MESSAGE_LAMBDA_FAILURE = "Lambda:Failure"
+LOG_MESSAGE_VALIDATION_FAILURE = "Validation:Failure"
+LOG_MESSAGE_VALIDATION_SUCCESS = "Validation:Success"
+LOG_MESSAGE_S3_BATCH_RESPONSE = "S3Batch:Response"
+LOG_MESSAGE_STEP_FUNCTION_RESPONSE = "StepFunction:Response"
```
l0b0, line 6: As elsewhere, it's probably best if these are two separate fields in the JSON rather than one, for easier search.

l0b0 (changes requested):

tests/test_validation_summary_logging.py
```diff
@@ -21,34 +21,34 @@ def should_log_event() -> None:
...
+    expected_log_call = call(LOG_MESSAGE_VALIDATION_SUCCESS, details=True)
     validation_results_model_mock.return_value.validation_outcome_index.count.return_value = 0
 
     with patch("geostore.validation_summary.task.LOGGER.debug") as logger_mock:
         # When
         task.lambda_handler(event, any_lambda_context())
 
         # Then
-        logger_mock.assert_any_call(expected_log)
+        logger_mock.assert_has_calls([expected_log_call])
```
l0b0, line 54: As discussed, this can be simplified to `logger_mock.assert_any_call(LOG_MESSAGE_VALIDATION_SUCCESS, details=True)`.

geostore/logging_keys.py
MitchellPaff: Could you please provide an example?  I don't quite understand.

MitchellPaff: is it something like.... 
log("Validation", result="Failure", foo=bar)
log("Validation", result="Success", foo=bar)

l0b0 (approved):
Fantastic!

PR #1259, opened 2021-12-03
feat: update dataset in place

MitchellPaff:


l0b0 (dismissed):

l0b0:
The missing cleanup code has caused us to [run out of our S3 bucket quota](https://github.com/linz/geostore/runs/4600560603?check_suite_focus=true) of [100](https://docs.aws.amazon.com/AmazonS3/latest/userguide/BucketRestrictions.html):

> ci1608864008-geostore | 10:23:32 PM | CREATE_FAILED        | AWS::S3::Bucket                  | api/cloudtrail/S3 (apicloudtrailS3644380FC) You have attempted to create more buckets than allowed (Service: Amazon S3; Status Code: 400; Error Code: TooManyBuckets; Request ID: 573EWBCVA0Q418X0; S3 Extended Request ID: a+cHTNTx7AU1yPTTtZK52jwzUUGqUFSYuod9TeNQLTNNtcrjzCdxxDxiFujnt0oGpoSl1/9Gf6Y=; Proxy: null)

I've deleted them (`./clear-s3-buckets.bash $(aws s3 ls | cut --delimiter=' ' --fields=3 | grep geostore)` and then `aws s3 ls | cut --delimiter=' ' --fields=3 | grep geostore | while read -r; do aws s3 rb "s3://${REPLY}"; done`), but please make sure that every test cleans up after itself.

l0b0 (dismissed):

l0b0 (approved):

PR #1286, opened 2021-12-21
fix: Bump Poetry installer checksum in production code

l0b0:
Works around the fact that the current production code relies on the
checksum of a third party file which has changed in the meantime.

See for example https://github.com/linz/geostore/runs/4525566154?check_suite_focus=true.

## Issues

Helps with #1287.

## Reference


billgeo (approved):
Seems harmless enough for me to approve. :)

PR #1294, opened 2021-12-22
fix: Match argument type to method signature

l0b0:
The value doesn't matter here. This prepares the code for mypy 0.930.

## Reference


billgeo (approved):
Thanks for the explanation. 👍

PR #1577, opened 2022-04-26
feat: Remove concept of Dataset STAC catalog in geostore

MitchellPaff:
Instead, new versions will be referenced directly in the root catalog. This is an intermediary step to allowing Datasets to be updated in place rather than having to create a new dataset version.

### DIFF
--- a/tests/test_processing_stack.py
+++ b/tests/test_processing_stack.py
@@ -260,17 +260,14 @@ def should_successfully_run_dataset_version_creation_process_with_multiple_asset
             )
             dataset_payload = load(dataset_response["Payload"])
 
-            assert dataset_payload.get(STATUS_CODE_KEY) == HTTPStatus.CREATED, dataset_payload
-
             dataset_id = dataset_payload[BODY_KEY][DATASET_ID_SHORT_KEY]
             dataset_prefix = f"{dataset_title}{DATASET_KEY_SEPARATOR}{dataset_id}"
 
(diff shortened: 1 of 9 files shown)


tests/general_generators.py
```diff
@@ -33,6 +33,10 @@ def any_past_datetime_string() -> str:
     return any_past_datetime().isoformat()
 
 
+def any_past_datetime_string_without_offset() -> str:
```
l0b0, line 36: https://github.com/linz/geostore/pull/1582#discussion_r863115795 is relevant here as well.

tests/test_processing_stack.py
```diff
@@ -312,48 +363,70 @@ def should_successfully_run_dataset_version_creation_process_with_multiple_asset
...
-                            STAC_REL_KEY: STAC_REL_SELF,
+                            STAC_TYPE_KEY: STAC_MEDIA_TYPE_JSON,
                         },
                     ],
                 }
 
             # Item contents
             imported_item_key = f"{dataset_version_prefix}{item_metadata_filename}"
+
+            # reintroduce when https://github.com/stac-utils/pystac/issues/801 is fixed
```
l0b0, line 386: Reintroduce what? Should that comment be just before line 405?

l0b0 (commented):
Nice! Just a couple minor things.

l0b0 (dismissed):

l0b0 (approved):

PR #1582, opened 2022-04-27
fix: multi asset processing test

MitchellPaff:
## Reference

### DIFF
--- a/tests/general_generators.py
+++ b/tests/general_generators.py
@@ -33,7 +33,7 @@ def any_past_datetime_string() -> str:
     return any_past_datetime().isoformat()
 
 
-def any_past_datetime_string_without_offset() -> str:
+def any_past_utc_datetime_string() -> str:
     return any_past_datetime().strftime("%Y-%m-%dT%H:%M:%SZ")
 
 
--- a/tests/general_generators.py
+++ b/tests/general_generators.py
@@ -33,7 +33,7 @@ def any_past_datetime_string() -> str:
     return any_past_datetime().isoformat()
 
 
-def any_past_datetime_string_without_tz() -> str:
+def any_past_datetime_string_without_offset() -> str:
     return any_past_datetime().strftime("%Y-%m-%dT%H:%M:%SZ")
 
 
--- a/tests/general_generators.py
+++ b/tests/general_generators.py
@@ -33,6 +33,10 @@ def any_past_datetime_string() -> str:
     return any_past_datetime().isoformat()
 
 
+def any_past_datetime_string_without_tz() -> str:
+    return any_past_datetime().strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
 def any_program_name() -> str:
     """Arbitrary-length string"""
     return random_string(20)
--- a/tests/test_processing_stack.py
+++ b/tests/test_processing_stack.py
@@ -407,34 +407,51 @@ def should_successfully_run_dataset_version_creation_process_with_multiple_asset
 
             # Item contents
             imported_item_key = f"{dataset_version_prefix}{item_metadata_filename}"
+
+            # reintroduce when https://github.com/stac-utils/pystac/issues/801 is fixed
             with subtests.test(msg="Imported item has relative keys"), smart_open.open(
                 f"{storage_bucket_prefix}{imported_item_key}", mode="rb"
             ) as imported_item_file:
-                assert load(imported_item_file) == {
-                    **deepcopy(MINIMAL_VALID_STAC_ITEM_OBJECT),
-                    STAC_ASSETS_KEY: {
-                        first_asset_name: {
-                            LINZ_STAC_CREATED_KEY: first_asset_created,
-                            LINZ_STAC_UPDATED_KEY: first_asset_updated,
-                            STAC_HREF_KEY: first_asset_filename,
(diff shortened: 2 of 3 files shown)


l0b0 (changes requested):
Looks like this test is broken.

MitchellPaff:
> Looks like this test is broken.

Yep - my purpose of raising this PR is to highlight that this test has been broken for a while. 
And to fix it before moving forward with the other changes.

tests/test_processing_stack.py
```diff
@@ -154,7 +168,56 @@ def should_successfully_run_dataset_version_creation_process_with_multiple_asset
...
+                        STAC_REL_KEY: STAC_REL_CHILD,
+                        STAC_HREF_KEY: f"./{dataset.dataset_prefix}/{CATALOG_FILENAME}",
+                        STAC_TITLE_KEY: dataset.title,
+                        STAC_TYPE_KEY: STAC_MEDIA_TYPE_JSON,
+                    },
+                ],
+            }
+        ),
+        bucket_name=Resource.STORAGE_BUCKET_NAME.resource_name,
+        key=CATALOG_FILENAME,
```
MitchellPaff, line 219: Will modify these to call Datasets Endpoint instead of creating the catalogs directly.

tests/general_generators.py
```diff
@@ -33,6 +33,10 @@ def any_past_datetime_string() -> str:
     return any_past_datetime().isoformat()
 
 
+def any_past_datetime_string_without_tz() -> str:
+    return any_past_datetime().strftime("%Y-%m-%dT%H:%M:%SZ")
```
l0b0, line 37: The "Z" part of the string is a timezone, so isn't this *with* time zone?

tests/test_processing_stack.py
```diff
@@ -263,31 +313,58 @@ def should_successfully_run_dataset_version_creation_process_with_multiple_asset
...
             with subtests.test(msg="Imported catalog has relative keys"), smart_open.open(
                 f"{storage_bucket_prefix}{imported_catalog_key}", mode="rb"
             ) as imported_catalog_file:
                 assert load(imported_catalog_file) == {
                     **deepcopy(MINIMAL_VALID_STAC_CATALOG_OBJECT),
+                    STAC_EXTENSIONS_KEY: [],  # pystac writes this field.
                     STAC_LINKS_KEY: [
                         {
-                            STAC_HREF_KEY: collection_metadata_filename,
+                            STAC_HREF_KEY: f"./{collection_metadata_filename}",
```
l0b0, line 353: This is not a valid S3 URL/path - S3 doesn't have relative paths. So I'm not sure it'd be appropriate to have the "./" part at the start here, as anyone trying to process the dataset would probably have to do manual path munging to get the right path. Ditto for `../../` below.

MitchellPaff: That depends on which layout strategy we intend to use.

In my mind we would be following the Relative Published strategy - where the root is an absolute s3 URL and everything else is a relative URL.

https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#relative-published-catalog

tests/general_generators.py
MitchellPaff: I think I meant no offset. Pystac was formatting datetimes like these and I was trying to align them.
would any_past_datetime_string_without_offset() be ok?

l0b0: I'd probably use "UTC" explicitly in the name rather than "without offset", so something like `any_past_utc_datetime_string`.

tests/test_processing_stack.py
```diff
@@ -220,17 +241,45 @@ def should_successfully_run_dataset_version_creation_process_with_multiple_asset
...
+                        BODY_KEY: {
+                            TITLE_KEY: dataset_title,
+                            DESCRIPTION_KEY: any_dataset_description(),
+                        },
+                    }
+                ).encode(),
+            )
+            dataset_payload = load(dataset_response["Payload"])
+
+            with subtests.test(msg="Dataset endpoint returns success"):
```
l0b0, line 263: A subtest means we want to continue executing the test even if the assertion fails. In this case that'll probably result in other spurious error messages, because it's very unlikely for the rest of the test to succeed if dataset creation failed. So we probably don't want to use a subtest here.

Alternatively we could consider this part of the test setup, and assume that the dataset creation functionality is being thoroughly tested elsewhere. We can then drop any assertions related to the setup, simply assuming that it'll work.

The same goes for the "Dataset catalog is created" test below.

tests/test_processing_stack.py
l0b0: I think that's a different thing. I'm just talking about how `./` and `../../` (as far as I know) won't do what you expect in an S3 link. For example, consider a user trying to crawl a dataset. They get to `s3://bucket/dataset/collection.json`, find a link `./foo.json`, and create a link to `foo.json` by concatenating `s3://bucket/dataset/` with `./foo.json`. I don't think that'll be a valid S3 link. They'll have to strip the `./` from the start before concatenating. Similarly, when dealing with links like `../catalog.json`, they'd have to strip a directory from the base URL for every `../` that they find in the link. On the other hand, since it seems S3 has no way to refer to a file in a parent directory with a relative link, we might as well pretend that we're on a "real" filesystem and use `./` and `../` for relative links.

l0b0 (approved):

l0b0 (approved):

PR #1610, opened 2022-05-05
feat: allow api role to be assumed by anyone in LINZ organization

MitchellPaff:
## Reference

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -117,6 +117,9 @@ Example of AWS service account authentication and authorization in to Geostore u
   export AWS_PROFILE=geostore-users
   ```
 
+Lambda and S3 access is also enabled by default for any AWS accounts in the LINZ organisation. This
+is by assuming the 'api-users' or 's3-users' role from the external AWS account.
+
 ## Use
 
 You can communicate with a Geostore instance using either the low-level application programming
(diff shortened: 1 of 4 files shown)


infrastructure/constructs/api.py
```diff
@@ -37,7 +37,7 @@ def __init__(  # pylint: disable=too-many-arguments,too-many-locals
             self,
             "api-users-role",
             role_name=Resource.API_USERS_ROLE_NAME.resource_name,
-            assumed_by=principal,
+            assumed_by=aws_iam.OrganizationPrincipal("o-g9kpx6ff4u"),
```
l0b0, line 40: Do we need to also replace the other use of `principal`, in the S3 users role? (At which point we can get rid of the `principal` parameter.)

billgeo: Just to clarify.

s3-users role should be able to be assumed by anyone at LINZ. So if you can achieve that through this same mechanism that's all good.

billgeo (commented):
Can you also look at the USAGE.md for any changes related to how people will assume roles?

USAGE.md
```diff
@@ -117,6 +117,9 @@ Example of AWS service account authentication and authorization in to Geostore u
   export AWS_PROFILE=geostore-users
   ```
 
+Lambda and S3 access is also enabled by default for any AWS accounts in the LINZ organisation. This
+is by assuming the 'api-users' or 's3_users' role from the external AWS account.
```
l0b0, line 121: Should `s3_users` be `s3-users`?

l0b0 (dismissed):

l0b0 (approved):

PR #1619, opened 2022-05-09
feat: allow partial dataset uploads

MitchellPaff:
## Reference


geostore/check_stac_metadata/task.py
```diff
@@ -64,21 +67,37 @@ def lambda_handler(event: JsonObject, _context: bytes) -> JsonObject:
 
     try:
         s3_client = get_s3_client_for_role(event[S3_ROLE_ARN_KEY])
+        geostore_s3_client = get_s3_client_for_role(get_param(ParameterName.S3_USERS_ROLE_ARN))
     except ClientError as error:
         LOGGER.warning(LOG_MESSAGE_LAMBDA_FAILURE, extra={"error": error})
         return {ERROR_MESSAGE_KEY: str(error)}
 
-    def s3_url_reader(url: str) -> StreamingBody:
+    def s3_url_reader(url: str, dataset_prefix: str) -> StreamingBody:
```
l0b0, line 75: Do we want to test this function in isolation? Then we shouldn't need more than a single test to ensure it's being used in the Lambda to ensure everything works.

l0b0 (changes requested):

geostore/check_files_checksums/task.py
```diff
@@ -18,6 +18,7 @@ def parse_arguments() -> Namespace:
     argument_parser = ArgumentParser()
     argument_parser.add_argument("--dataset-id", required=True)
     argument_parser.add_argument("--version-id", required=True)
```
l0b0, line 20: As discussed, this argument can probably go now.

geostore/check_files_checksums/task.py
```diff
@@ -18,6 +18,7 @@ def parse_arguments() -> Namespace:
...
+    argument_parser.add_argument("--dataset-prefix", required=True)
```
l0b0, line 21: There doesn't seem to be any tests with a *specific* dataset prefix value.

geostore/check_files_checksums/utils.py
```diff
@@ -75,30 +74,38 @@ def validate(self, hash_key: str, range_key: str) -> None:
         else:
             self.logger.info(LOG_MESSAGE_VALIDATION_COMPLETE, extra={"outcome": Outcome.PASSED})
             self.validation_result_factory.save(item.url, Check.CHECKSUM, ValidationResult.PASSED)
+            if not file_in_staging:
+                item.delete()
```
l0b0, line 78: This doesn't seem right. The file is mentioned in the staging metadata, so it should stay unchanged.

geostore/check_stac_metadata/utils.py
```diff
@@ -147,7 +151,17 @@ def process_assets(self) -> None:
 
     def validate(self, url: str) -> None:  # pylint: disable=too-complex
         self.traversed_urls.append(url)
-        object_json = self.get_object(url)
+        s3_response = self.get_object(url)
+        try:
+            object_json: JsonObject = load(
+                s3_response.response,
+                object_pairs_hook=self.duplicate_object_names_report_builder(url),
```
l0b0, line 158: Should this maybe be a separate PR? I basically forgot to include this functionality in at least one place (this one) where we parse metadata.

geostore/check.py
```diff
@@ -5,8 +5,9 @@ class Check(Enum):
     ASSETS_IN_DATASET = "assets in dataset"
     CHECKSUM = "checksum"
     DUPLICATE_OBJECT_KEY = "duplicate asset name"
+    FILE_NOT_FOUND = "file not found in staging or storage"
```
l0b0, line 8: I just realised these errors should probably be fields on the relevant exception classes. For example, this one should be on the `DatasetFileNotFoundError` class. But that's something for another PR.

tests/test_s3_utils.py
```diff
@@ -46,7 +59,77 @@ def should_successfully_get_object_from_staging_bucket() -> None:
...
-        s3_url_reader = get_s3_url_reader(get_s3_role_arn())
-        json_object = load(s3_url_reader(collection_metadata_url))
+        s3_url_reader = get_s3_url_reader(get_s3_role_arn(), any_dataset_prefix(), get_log())
+        json_object = load(s3_url_reader(collection_metadata_url).response["Body"])
+
+        assert json_object == collection_dict
+
+
+@mark.infrastructure
+def should_get_object_from_storage_bucket_when_staging_bucket_fails() -> None:
```
l0b0, line 69: "when_staging_bucket_fails" could maybe be clarified as "when_not_in_staging_bucket"?

tests/test_s3_utils.py
```diff
@@ -46,7 +59,77 @@ def should_successfully_get_object_from_staging_bucket() -> None:
...
+    key_prefix = any_safe_file_path()
+    metadata_url_prefix = (
+        f"{S3_URL_PREFIX}{Resource.STAGING_BUCKET_NAME.resource_name}/{key_prefix}"
+    )
+    collection_metadata_filename = any_safe_filename()
+    collection_metadata_url = f"{metadata_url_prefix}/{collection_metadata_filename}"
+
+    collection_dict = {
```
l0b0, line 77: It looks like the contents of this collection doesn't actually matter, so I think it could be simplified to just `deepcopy(MINIMAL_VALID_STAC_COLLECTION_OBJECT)` or something like `any_json_object()`.

geostore/check_files_checksums/task.py
MitchellPaff: This is still required to build the hash key for querying the processing assets table.

geostore/check.py
MitchellPaff: Do you mean in addition to here also?

Most of these "Checks" don't correspond to a defined Exception.

l0b0 (dismissed):

l0b0 (dismissed):

PR #1727, opened 2022-06-09
fix: Remove "prod-" prefix from production Lambda endpoints

l0b0:
This is one of the few differences between production and all other
environments, so it was never tested.

## Post-deployment steps

- [ ] Verify end user [Lambda names](https://ap-southeast-2.console.aws.amazon.com/lambda/home?region=ap-southeast-2#/functions)

## Issues

Closes https://github.com/linz/geostore/issues/1726.

## Reference

### DIFF
--- a/infrastructure/constructs/lambda_endpoint.py
+++ b/infrastructure/constructs/lambda_endpoint.py
@@ -25,8 +25,8 @@ class LambdaEndpoint(aws_lambda.Function):
     ):
         super().__init__(
             scope,
-            f"{env_name}-{construct_id}-function",
-            function_name=f"{env_name}-{construct_id}",
+            f"{construct_id}-function",
+            function_name=construct_id,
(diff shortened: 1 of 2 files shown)


infrastructure/constructs/lambda_endpoint.py
```diff
@@ -25,8 +25,8 @@ def __init__(
     ):
         super().__init__(
             scope,
-            f"{env_name}-{construct_id}-function",
-            function_name=f"{env_name}-{construct_id}",
+            f"{construct_id}-function",
```
billgeo, line 28: Does this mean the function name will be identical in nonprod and prod?

l0b0: No, the construct ID is the thing which has the "env-" prefix for non-prod environments. So the names should only be different in prod.

billgeo (approved):

MitchellPaff (approved):

PR #1732, opened 2022-06-09
feat: add new filename field to processing assets table

MitchellPaff:
## Reference

### DIFF
--- a/geostore/check_stac_metadata/utils.py
+++ b/geostore/check_stac_metadata/utils.py
@@ -1,7 +1,7 @@
 from functools import lru_cache
 from json import JSONDecodeError, load
 from logging import Logger
-from os.path import dirname
+from os.path import basename, dirname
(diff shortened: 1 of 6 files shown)


geostore/check_stac_metadata/utils.py
```diff
@@ -137,6 +137,7 @@ def process_metadata(self) -> None:
                 hash_key=self.hash_key,
                 range_key=f"{ProcessingAssetType.METADATA.value}{DB_KEY_SEPARATOR}{index}",
                 url=metadata_file[PROCESSING_ASSET_URL_KEY],
+                filename=basename(str(metadata_file[PROCESSING_ASSET_URL_KEY])),
```
l0b0, line 140: You can avoid the `str()` conversion by pulling out the value and doing `assert isinstance(asset_url, str)`. Assertions are stripped out when optimising Python, so it shouldn't affect production speed.

l0b0 (approved):

PR #1745, opened 2022-06-13
refactor: checksum code refactoring

MitchellPaff:
## Reference

### DIFF
--- a/geostore/check_files_checksums/utils.py
+++ b/geostore/check_files_checksums/utils.py
@@ -3,6 +3,7 @@ from os import environ
 from typing import Callable
 
 from botocore.exceptions import ClientError
+from botocore.response import StreamingBody
 from multihash import FUNCS, decode
 
 from ..api_keys import MESSAGE_KEY
@@ -19,14 +20,7 @@ from ..validation_results_model import ValidationResult, ValidationResultFactory
 ARRAY_INDEX_VARIABLE_NAME = "AWS_BATCH_JOB_ARRAY_INDEX"
 
 
-class ChecksumMismatchError(Exception):
-    def __init__(self, actual_hex_digest: str):
-        super().__init__()
-
-        self.actual_hex_digest = actual_hex_digest
-
-
-class ChecksumValidator:
+class ChecksumUtils:
     def __init__(
         self,
         processing_assets_table_name: str,
@@ -48,9 +42,10 @@ class ChecksumValidator:
             LOG_MESSAGE_VALIDATION_COMPLETE, extra={"outcome": Outcome.FAILED, "error": content}
         )
 
-    def validate(self, hash_key: str, range_key: str) -> None:
+    def run(self, hash_key: str, range_key: str) -> None:
+
         try:
-            item = self.processing_assets_model.get(hash_key, range_key=range_key)
+            processing_item = self.processing_assets_model.get(hash_key, range_key=range_key)
         except self.processing_assets_model.DoesNotExist:
             self.log_failure(
                 {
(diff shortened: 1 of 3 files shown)


geostore/check_files_checksums/utils.py
```diff
@@ -60,28 +55,44 @@ def validate(self, hash_key: str, range_key: str) -> None:
...
             self.logger.info(LOG_MESSAGE_VALIDATION_COMPLETE, extra={"outcome": Outcome.PASSED})
-            self.validation_result_factory.save(item.url, Check.CHECKSUM, ValidationResult.PASSED)
-            item.update(
-                actions=[self.processing_assets_model.exists_in_staging.set(file_in_staging)]
-            )
-
-    def validate_url_multihash(self, url: str, hex_multihash: str) -> bool:
+            self.validation_result_factory.save(url, Check.CHECKSUM, ValidationResult.PASSED)
 
+    def get_s3_object(self, url: str) -> GeostoreS3Response:
```
l0b0, line 93: I like this change - everything related to just getting the object in one place.

geostore/check_files_checksums/utils.py
```diff
@@ -19,14 +20,7 @@
 ARRAY_INDEX_VARIABLE_NAME = "AWS_BATCH_JOB_ARRAY_INDEX"
 
 
-class ChecksumMismatchError(Exception):
```
l0b0, line 22: It was good to get rid of this. Raising and catching an exception within a single piece of code was definitely a code smell.

geostore/check_files_checksums/utils.py
```diff
@@ -48,9 +42,10 @@ def log_failure(self, content: JsonObject) -> None:
             LOG_MESSAGE_VALIDATION_COMPLETE, extra={"outcome": Outcome.FAILED, "error": content}
         )
 
-    def validate(self, hash_key: str, range_key: str) -> None:
+    def run(self, hash_key: str, range_key: str) -> None:
```
l0b0, line 45: `run` is a very generic name, which I'd normally reserve for methods which do nothing but dispatch, like `main`. This method fetches some asset metadata from DynamoDB and then validates the asset multihash. Maybe a name like `validate_asset_multihash`?

geostore/check_files_checksums/utils.py
```diff
@@ -60,28 +55,44 @@ def validate(self, hash_key: str, range_key: str) -> None:
...
+        processing_item.update(
+            actions=[
+                self.processing_assets_model.exists_in_staging.set(s3_response.file_in_staging)
+            ]
+        )
+
+    def validate_url_multihash(
+        self, url: str, hex_multihash: str, s3_file_object: StreamingBody
+    ) -> None:
+        checksum_function_code = int(hex_multihash[:2], 16)
```
l0b0, line 73: I was able to pull out a function to get the digest which might work for us:

```python
if TYPE_CHECKING:
    from hashlib import _Hash

def get_multihash_digest(digest_algorithm_code: int, body: StreamingBody) -> bytes:
    hash_object: "_Hash" = FUNCS[digest_algorithm_code]()

    for chunk in body.iter_chunks(chunk_size=CHUNK_SIZE):
        hash_object.update(chunk)

    return hash_object.digest()
```

geostore/check_files_checksums/utils.py
MitchellPaff: I was trying to be generic, because after the changes on following pull request this method will also update processing assets table with 'exist_in_staging' field and 'replaced_in_new_version' field. Both of which don't relate to validation.

l0b0 (approved):

PR #1749, opened 2022-06-14
feat: mark files as replaced

MitchellPaff:
## Reference


geostore/check_stac_metadata/utils.py
```diff
@@ -150,7 +156,7 @@ def process_assets(self) -> None:
             assert isinstance(asset_url, str)
 
             self.processing_assets_model(
-                hash_key=self.hash_key,
+                hash_key=get_hash_key(self.dataset_id, self.new_version_id),
```
l0b0, line 159: Looks like `get_hash_key` is called for each metadata file and asset. We probably want to call it in the constructor instead.

geostore/step_function.py
```diff
@@ -180,3 +181,29 @@ def get_hash_key(dataset_id: str, dataset_version_id: str) -> str:
...
+    items = processing_assets_model.query(
+        get_hash_key(dataset_id, current_version_id),
+        range_key_condition=processing_assets_model.sk.startswith(
+            f"{processing_asset_type.value}{DB_KEY_SEPARATOR}"
+        ),
+        filter_condition=(processing_assets_model.filename == filename),
+        consistent_read=True,
+    )
+
+    assert items.total_count > 1, "More than 1 file returned with the same filename"
```
l0b0, line 205: Did you mean to `assert items.total_count == 1`?

Also, the assertion message should probably include the relevant parameters to debug it, basically the dataset ID, version ID and filename.

geostore/step_function.py
```diff
@@ -180,3 +181,29 @@ def get_hash_key(dataset_id: str, dataset_version_id: str) -> str:
...
+
+    if items.total_count == 1:
```
l0b0, line 207: This shouldn't be necessary with the assertion above.

geostore/step_function.py
MitchellPaff: No, I meant to assert that total is  <2. 

It's a valid case that this query returns 0 - this is the case when this file is new in this version.

billgeo (approved):
Commits so far look good. But looks like a couple of tests are failing?

billgeo (approved):

tests/test_check_stac_metadata.py
```diff
@@ -694,6 +698,72 @@ def should_successfully_validate_partially_uploaded_dataset(subtests: SubTests)
                     )
 
 
+@mark.timeout(timedelta(minutes=20).total_seconds())
+@mark.infrastructure
+def should_mark_asset_as_replaced_in_new_version() -> None:
```
l0b0, line 703: There are two tests with the exact same name. Based on #1779 I assume that this test was not meant to be here.

l0b0 (dismissed):

l0b0 (approved):

PR #1822, opened 2022-07-06
feat: Switching to OIDC connect

Jimlinz:
Draft PR needed to trigger workflow on GitHub actions. This is needed for testing.

### DIFF
--- a/.github/workflows/cdk-bootstrap.yml
+++ b/.github/workflows/cdk-bootstrap.yml
@@ -8,6 +8,9 @@ on:
 jobs:
   boostrap-ci:
     runs-on: ubuntu-20.04
+    permissions:
+      id-token: write
+      contents: read
     steps:
       - name: Check out repository
         uses: actions/checkout@v3
@@ -70,11 +73,9 @@ jobs:
       - name: (CI) Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1.6.1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_CI }}
+          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_CI_OIDC }}
 
       - name: Bootstrap CDK
         run: |
@@ -82,6 +83,9 @@ jobs:
 
   boostrap-nonprod:
     runs-on: ubuntu-20.04
+    permissions:
+      id-token: write
+      contents: read
     steps:
       - name: Check out repository
         uses: actions/checkout@v3
@@ -144,8 +148,6 @@ jobs:
       - name: (Non-Prod) Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1.6.1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ap-southeast-2
           mask-aws-account-id: true
           role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_NON_PROD }}
@@ -156,6 +158,9 @@ jobs:
 
   boostrap-prod:
     runs-on: ubuntu-20.04
+    permissions:
+      id-token: write
+      contents: read
     steps:
       - name: Check out repository
         uses: actions/checkout@v3
@@ -218,8 +223,6 @@ jobs:
       - name: (Prod) Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1.6.1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ap-southeast-2
           mask-aws-account-id: true
           role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_PROD }}
--- a/.github/workflows/cdk-bootstrap.yml
+++ b/.github/workflows/cdk-bootstrap.yml
@@ -75,7 +75,7 @@ jobs:
         with:
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: arn:aws:iam::586981104868:role/AWS_ASSUME_ROLE_CI_OIDC
+          role-to-assume: arn:aws:iam::586981104868:role/CiOidc
 
       - name: Bootstrap CDK
         run: |
@@ -150,7 +150,7 @@ jobs:
         with:
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: arn:aws:iam::632223577832:role/AWS_ASSUME_ROLE_NON_PROD_OIDC
+          role-to-assume: arn:aws:iam::632223577832:role/NonProdOidc
 
       - name: Bootstrap CDK
         run: |
@@ -225,7 +225,7 @@ jobs:
         with:
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: arn:aws:iam::715898075157:role/AWS_ASSUME_ROLE_PROD_OIDC
+          role-to-assume: arn:aws:iam::715898075157:role/ProdOidc
 
       - name: Bootstrap CDK
         run: |
--- a/.github/workflows/cdk-bootstrap.yml
+++ b/.github/workflows/cdk-bootstrap.yml
@@ -75,7 +75,7 @@ jobs:
         with:
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_CI_OIDC }}
+          role-to-assume: arn:aws:iam::586981104868:role/AWS_ASSUME_ROLE_CI_OIDC
 
       - name: Bootstrap CDK
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -112,11 +115,22 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1.6.1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_CI }}
+          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_CI_OIDC }}
```
l0b0, line 120: Role names should not be secrets. We should probably hard-code this to make it easier to debug role issues.

.github/workflows/test.yml
```diff
@@ -112,11 +115,22 @@ jobs:
...
+
+      - name: Set up Boto3 with AssumeRoleWithWebIdentity for oidc connectivity
+        run: |
+          export AWS_ROLE_ARN=${{ secrets.AWS_ASSUME_ROLE_CI_OIDC }}
+          export AWS_WEB_IDENTITY_TOKEN_FILE=/tmp/awscreds
+          export AWS_DEFAULT_REGION=ap-southeast-2
```
l0b0, line 126: `export`ing these should be unnecessary here, because that just makes the variable available *within* other processes. It looks like all references to these variables in this step are explicit.

.github/workflows/test.yml
```diff
@@ -112,11 +115,22 @@ jobs:
...
+
+          echo AWS_WEB_IDENTITY_TOKEN_FILE=$AWS_WEB_IDENTITY_TOKEN_FILE >> $GITHUB_ENV
+          echo AWS_ROLE_ARN=$AWS_ROLE_ARN >> $GITHUB_ENV
+          echo AWS_DEFAULT_REGION=$AWS_DEFAULT_REGION >> $GITHUB_ENV
```
l0b0, line 130: Since these are static values they might as well be set in the `env` for the whole job.

infrastructure/application_stack.py
```diff
@@ -33,8 +33,20 @@ def __init__(self, scope: constructs.Construct, stack_id: str) -> None:
...
         else:
-            principal = aws_iam.AccountPrincipal(
-                account_id=aws_iam.AccountRootPrincipal().account_id
+            open_id_connect_provider_arn = (
+                f"arn:aws:iam::"
+                f"{aws_iam.AccountRootPrincipal().account_id}"
+                f":oidc-provider/token.actions.githubusercontent.com"
+            )
+
+            principal = aws_iam.WebIdentityPrincipal(
```
l0b0, line 42: You might need a `CompositePrincipal` which also includes the original account root principal here to fix the S3 role assumption.

l0b0: Commit afbc25df6d00a2e9861c6fb80730a30258d18153 seems to have indeed fixed this issue.

.github/workflows/test.yml
linz-jim: Arh. Yep. Good point. This was mostly me testing, trying to work out if any other processes were consuming these variables. I have removed and moved them to env as per your other suggestion.

linz-jim:
### Switching to OIDC

This change moves **Geostore** from accessing AWS resources using long-lived AWS credentials stored in GitHub secrets to _OpenID Connect_ (OIDC). It eliminates the need to rotate AWS keys on a periodic basis.

Setting up [oidc-provider](https://github.com/linz/oidc-provider) is a prerequisite to this change. The oidc-provider stack only needs to be deployed once per AWS account. It sets out the necessary permissions for GitHub actions to access AWS resources. Restrictions can be put in place by limiting access to specific repo and branch. The oidc-provider stack also creates a role that is assumed by GitHub actions. Within the context of this Geostore repo, the following roles need to be created in their respective AWS account prior to the change: 

- AWS_ASSUME_ROLE_NON_PROD_OIDC
- AWS_ASSUME_ROLE_PROD_OIDC
- AWS_ASSUME_ROLE_CI_OIDC

AWS credentials should be removed from GitHub secrets post change.

.github/workflows/cdk-bootstrap.yml
```diff
@@ -70,18 +73,19 @@ jobs:
       - name: (CI) Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1.6.1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_CI }}
+          role-to-assume: arn:aws:iam::586981104868:role/AWS_ASSUME_ROLE_CI_OIDC
```
l0b0, line 78: Role names are normally `CamelCase`. Also, the words "AWS" and "role" in the AWS role name are redundant. Maybe something like `OidcCi` or `CiOidc`?

.github/workflows/cdk-bootstrap.yml
```diff
@@ -144,18 +148,19 @@ jobs:
       - name: (Non-Prod) Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v1.6.1
         with:
-          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_NON_PROD }}
+          role-to-assume: arn:aws:iam::632223577832:role/AWS_ASSUME_ROLE_NON_PROD_OIDC
```
l0b0, line 153: Is there some way to de-duplicate this between jobs and workflows? I don't think GitHub workflows have a way to include a file with global variables, but I might be wrong.

.github/workflows/test.yml
```diff
@@ -3,6 +3,11 @@ on:
   pull_request:
     types: [opened, synchronize]
 
+env:
+  AWS_WEB_IDENTITY_TOKEN_FILE: /tmp/awscreds
```
l0b0, line 7: Skimming some [AWS documentation around this subject](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) it looks like a path like `/var/run/secrets/eks.amazonaws.com/[role name]/token` might be more idiomatic. What do you think?

l0b0:
By the way, the title of the PR ends up as the title of the commit created when it's merged, so the PR name should be formatted like a commit message. In this case, that'd be "feat: OIDC connect"

.github/workflows/cdk-bootstrap.yml
linz-jim: It's probably tidier ya. I copied the role from Github Secrets without much thought. Have fixed.

.github/workflows/cdk-bootstrap.yml
linz-jim: An idea that **might** work 

We could create a .env file and populate it with the necessary values (e.g. CiOidc=arn:aws..... NonProdOidc=arn:aws), and check this in to the .github/workflows directory

At the start of the workflow after repo checkout, we can run 
cat .env >> $GITHUB_ENV

Worth trying out?

l0b0: Sounds good!

linz-jim: Updated in [00f8a46](https://github.com/linz/geostore/pull/1822/commits/00f8a46a054dd4c80906ac00fa73f511f3927a00).

.github/workflows/test.yml
Jimlinz: The directory `/var/run/secrets/eks.amazonaws.com/[role name]/` need to exists first before we can store the token there no? Otherwise we will just get a `No such file or directory` error message?

l0b0: I don't know what CDK does with that path, but it's possible it'll create the directory before attempting to create the file. If so, changing this should just work™. If not, we could either create the directory ourselves or use a more obvious `/tmp` filename based on that path, like `/tmp/aws-eks-[role-name]-token`.

l0b0 (dismissed):

.github/workflows/test.yml
Jimlinz: Gave it a go in [b06747e](https://github.com/linz/geostore/commit/b06747ed9cf31715217c80cc611d10adffd9f3c7) and it didn't work. Have opted for a more obvious filename in tmp. Thought about creating the path but my preference would be to keep the workflow steps simple. I see the PR has been approved, though I am more than happy to revisit and tidy up later.

.github/workflows/test.yml
```diff
@@ -109,14 +112,20 @@ jobs:
...
-          aws-region: ap-southeast-2
+          aws-region: ${{ env.AWS_DEFAULT_REGION }}
           mask-aws-account-id: true
-          role-to-assume: ${{ secrets.AWS_ASSUME_ROLE_CI }}
+          role-to-assume: ${{ env.CiOidc }}
+
+      - name: Set up Boto3 with AssumeRoleWithWebIdentity for oidc connectivity
+        run: |
+          curl -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL" \
+          | jq -r '.value' > /tmp/aws-eks-CiOidc-token
```
l0b0, line 128: This path isn't referenced anywhere in this PR - that seems like an oversight.

Jimlinz: Huh. You are right! Turns out I added this in https://github.com/linz/geostore/commit/54dd65527ace55b9d73b853b0fa6a0fd5ed38e88 in an attempt to use `AWS_WEB_IDENTITY_TOKEN_FILE` to allow s3 role assumption. This was fixed in [afbc25d](https://github.com/linz/geostore/pull/1822/commits/afbc25df6d00a2e9861c6fb80730a30258d18153), but the original addition was not removed.

l0b0 (approved):

PR #1843, opened 2022-07-12
docs: Recommend portable Markdown syntax

l0b0:
## Reference

### DIFF
--- a/CODING.md
+++ b/CODING.md
@@ -289,6 +289,29 @@ validated automatically.
    output) to enable more user-friendly formatting such as titles for single values, table headings
    and aligned table columns, and pretty-printed JSON, any of which could be coloured.
 
+#### Documentation
+
+-  We should try to use portable
+   [Markdown syntax](https://daringfireball.net/projects/markdown/syntax). This isn't always easy -
+   even mainstream renderers like GitHub, Stack Overflow, and JetBrains IDEs disagree about subtle
+   details, which are usually only discoverable by trial and error. Rationale: As an open source
+   project, we should focus on portability where it's cheap, and only add non-portable features when
+   it's really valuable. Basically we need to balance readability and ease of editing across
+   different platforms.
+
+   Bad example: ASCII art is generally neither self-documenting nor reproducible. That is, if you
+   wanted to change a single character in a non-trivial ASCII art diagram you have several problems.
+   Which ASCII art editor was used? Since it's not self-documenting there's no way to know unless
+   you use precious documentation space for this meta-documentation. And what was the source of the
+   ASCII art? Many of them let you create a diagram which is then exported to ASCII art, but going
+   the other way is generally not possible. Bit rot is also a concern - what if the editor has
+   changed fundamentally or no longer exists? And finally, a lot of editors can't be trivially
+   automated, so any kind of updating might require manual work every time.
+
+   Good example: A [DOT file](https://graphviz.org/doc/info/lang.html) is a relatively simple
+   format, can be converted into various image formats, and the conversion is trivial to automate.
+   Including the result of such a conversion in Markdown is also trivial.
+
 #### Code reuse
 
 -  Group code by relatedness, not by reuse. Basically, putting a bunch of stuff in a single file


CODING.md
```diff
@@ -289,6 +289,29 @@ validated automatically.
    output) to enable more user-friendly formatting such as titles for single values, table headings
    and aligned table columns, and pretty-printed JSON, any of which could be coloured.
 
+#### Documentation
+
+-  We should try to use portable
+   [Markdown syntax](https://daringfireball.net/projects/markdown/syntax). This isn't always easy -
```
linz-jim, line 295: There are quite a few Markdown guides out there. Most of them cover the basics. Are we keeping to the guide you've provided? GitHub docs used to contain a section on Markdown IIRC but they have now changed this to "Basic writing and formatting syntax". There is also [Markdown guide](https://www.markdownguide.org/).

Probably a moot point since the basics are the same unless one of us veer off and try to do something funky like adding colours or emoji (please don't do this, though GitHub allows it).

l0b0: This syntax page is the original Markdown reference. Hence the reference to it being portable.

billgeo: [The markdown syntax ](https://www.markdownguide.org/basic-syntax/) looks slightly more user friendly to me. It has a better layout and easier to jump to the bit you need. Also seems to have a bit more info about portability. Such as [this](https://www.markdownguide.org/basic-syntax/#italic-best-practices). So if it's acceptable as a guide to keep it portable we could reference that one here?

l0b0: The original Markdown spec was underspecified, so implementers have made different choices for how to render input. For example, nested lists are a pain. The standard says "List markers typically start at the left margin, but may be indented by up to three spaces." Obviously this is talking about top-level lists, not nested ones. So three spaces followed by a list marker should still be a top-level list element. But many implementers seem to treat three (sometimes even two) spaces as meaning a *sub-list* element. So according to the standard, this code should be two entries in the same top-level list:

```markdown
* foo
   * bar
```

However, many renderers, including GitHub, will treat the second line as being a sub-list entry.

Most implementers have also extended the standard (tables, footnotes, etc.), but the original page was never updated to include these extensions. So the best we can do is avoid extensions, stick to the original standard where possible, and find hacks to appease all relevant renderers to write "as portable as possible" Markdown. That's why I linked to the original standard - any other guides are likely to introduce extensions and to decide on ambiguous or missing standardisation in a way that's guaranteed to work in only a subset of renderers (because renderers have made fundamentally incompatible decisions).

The guidelines I use are based on experience with different renderers, not a specific standard, so take them as you will, but here are the ones which come to mind. In very general terms:

- Use a lot of vertical space. Top-level headings (except at the very top of the file), paragraphs, lists, and code blocks should have an empty line above and below them.
- If in doubt, test with non-trivial inputs. For example, try creating a four-level-deep numbered list with multiple entries on each level, and check which indentation works with the relevant renderers. As far as I can tell, three spaces seems to work with most renderers, even though that's wrong as seen above. Also, if you're unsure whether an unordered list is actually a single list, simply replace the list markers with `1.` and see whether the numbers all line up. If they restart from 1 you've probably added some unnecessary newline.
- Use `1.` as the numbered list marker. Anything else results in an unnecessarily big diff when inserting/removing in the middle of the list, and lots of unnecessary busywork to keep the numbers correct.
- Use at most a single H1 (`# Foo`) at the top of each file. If a file needs more than one, it should probably be split.
- Mark code blocks with the language name, such as <code>```bash</code>, to help ensure correct formatting. Language names unfortunately are not standardised, and simply map to internal names in renderers. For example, some renderers want you to use a misleading "shell" name, as if that were a single language.
- Escaping is a giant mess. If you need to do a bunch of escaping, wrap it in the equivalent HTML instead.
- Most extensions are highly non-portable; don't use them.
- Some extensions, such as tables, result in horrible Markdown. Try to make sure that the Markdown is just as readable as the rendered result.

billgeo: Not suggesting you write about this any more than you have already. Just thought https://www.markdownguide.org/ looked easier to use and seemed to have a good handle on portability. But if you think https://daringfireball.net/projects/markdown/syntax is a better guide for ensuring portability, then all good.

billgeo: It's fine anyway. Approving. 😄

billgeo (approved):

PR #1859, opened 2022-07-18
refactor: Speed up CDK

Jimlinz:
This pull request optimises Geostore CDK deployment time in GitHub Actions workflow by calling `aws cloudformation delete-stack` rather than `cdk destroy`

Running `cdk destroy` causes cdk to parse through the stack again before attempting to tear it down. It waits for the stack to be completely destroyed before finishing the workflow. ` aws cloudformation delete-stack` is a simple api call that doesn't wait for a response. It allows the call to be made to aws to terminate the stack. This shaves about ~5 to 6 minutes off the workflow runtime.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -163,7 +163,7 @@ jobs:
 
       - name: Destroy AWS stacks used for testing
         run: |
-          poetry run cdk destroy --force --all
+          aws cloudformation delete-stack --stack-name ci${GITHUB_RUN_ID}-geostore
         if: always()
 
   build-nix-shell:


.github/workflows/test.yml
```diff
@@ -163,7 +163,11 @@ jobs:
 
       - name: Destroy AWS stacks used for testing
         run: |
-          poetry run cdk destroy --force --all
+          cdk --app cdk.out ls | while read line ; do
```
l0b0, line 166: This lists the OIDC stack as well, which we don't want to destroy. You can avoid that by adding `'*-geostore'` to print only the Geostore stack.

.github/workflows/test.yml
```diff
@@ -163,7 +163,11 @@ jobs:
...
+             if [ "$line" != "" ] ; then
```
l0b0, line 167: As far as I can tell this isn't necessary. The empty line printed by `cdk` is on standard error, so it won't be sent to the `read` command.

.github/workflows/test.yml
```diff
@@ -163,7 +163,11 @@ jobs:
...
+                aws cloudformation delete-stack --stack-name $line
```
l0b0, line 168: `$line` should be quoted to avoid any weirdness when substituting it, as in `"$line"`.

cdk.context.json
```diff
@@ -30,37 +30,6 @@
       }
     ]
   },
-  "vpc-provider:account=368354993240:filter.tag:ApplicationLayer=networking:filter.tag:ApplicationName=geostore:region=ap-southeast-2:returnAsymmetricSubnets=true": {
```
l0b0, line 33: I would probably put the changes to this file in a separate PR, since commits are squashed when merging. That way, if we ever need to undo it's a trivial, single-commit revert.

Jimlinz: Have moved this to #1887

.github/workflows/test.yml
Jimlinz: Have removed loop altogether to keep things simple for now. Happy to revisit should we introduce another stack into the mix in the future. Parsing through `cdk.out` is much faster than running `cdk ls`, though there is still some overhead (few seconds lead time) to get the actual stack names listed.

l0b0 (approved):

PR #1896, opened 2022-07-29
fix: Skip test for .md changes

Jimlinz:
No need to trigger long-running test workflow for .md changes.

[Reference](https://stackoverflow.com/questions/62968897/is-it-possible-to-not-run-github-action-for-readme-updates)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,8 @@ name: Pull request
 on:
   pull_request:
     types: [opened, synchronize]
+    paths-ignore:
+      - '**.md'
 
 jobs:
   linter:


.github/workflows/test.yml
```diff
@@ -2,6 +2,8 @@ name: Pull request
 on:
   pull_request:
     types: [opened, synchronize]
+    paths-ignore:
+      - '**.md'
```
l0b0, line 6: We *do* want to run the linter when Markdown files change, so it might be easiest to split the linting step into a separate workflow without this option.

l0b0: Also, don't think it's possible to set the [branch protection rules](https://github.com/linz/geostore/settings/branch_protection_rules/17596251) to deal with the new situation. You probably need to add a [`finalize` step](https://github.com/linz/bde-processor-deployment/blob/81e5e5d3a25b2430c1e5b13569a4da82b1f47b1d/.github/workflows/test.yml#L276-L287) and use [`allowed-skips`](https://github.com/re-actors/alls-green) for the tests.

Jimlinz: Forgotten about `.md` linting. Probably jumped the gun a bit wondering why we trigger long running (>30 min) tests deploying infrastructure to AWS for simple `.md` change. Linting is needed yep. Will have a rethink.

Jimlinz:
This [repo](https://github.com/dorny/paths-filter) is the closest thing I found that allows GitHub actions to conditionally run based on files modified by PR.

Though I am starting to wonder if this is more hassle than it is worth since `.md` files rarely get updated....

Jimlinz:
This has been opened for some time. Closing and will revisit if / when the need arise.

Issue #1818, opened 2022-07-04
Bug: New version of every metadata file when a new dataset is added


billgeo:
## Bug Description

Every time a new dataset is added, it creates a new version of every other metadata file in the s3 bucket. This may be expected in the way we use pystac, but it could potentially add up to Terabytes of data if we have a high number of datasets with a high number of metadata files? It could also lead to peformance issues?

First need to investigate and find best options to resolve this.
Could be related to this? https://github.com/stac-utils/pystac/issues/90

See here for example of many file versions:

![Screenshot from 2022-07-05 08-24-26.png](https://images.zenhubusercontent.com/5ab31db94b5806bc2bcbe143/f731b2c3-4eb2-487b-bfa1-e099f55fd2d6)

## How to Reproduce

1. Add a new 'dataset'
1. Add a new 'dataset version'
1. Add a second 'dataset version'
1. See new versions for metadata files in other sibling datasets

What did you expect to happen? 

What actually happened? 

## Software Context

Operating system: 

Environment: 

Relevant software versions:

-  AWS CLI: 
-  Poetry: 

#### Definition of Done

-  [ ] This bug is **done**:
   -  [ ] Bug resolved to **user's** satisfaction
   -  [ ] Automated tests are passing
   -  [ ] Code is peer reviewed and pushed to master
   -  [ ] Deployed successfully to test environment
   -  [ ] Checked against
          [CODING guidelines](https://github.com/linz/geostore/blob/master/CODING.md)
   -  [ ] Relevant new tasks are added to backlog and communicated to the team
   -  [ ] Important decisions recorded in the issue ticket
   -  [ ] Readme/Changelog/Diagrams are updated
   -  [ ] Product Owner has approved as complete
   -  [ ] No regression to functional or
          [non-functional](https://github.com/linz/geostore/blob/master/.github/ISSUE_TEMPLATE/user_story.md)
          requirements

billgeo:
Tested this and it's working as expected. Collection.json has a new version, but the child items don't.  Thanks @Jimlinz !

FYI - I also created a followup issue https://github.com/linz/geostore/issues/2022 to undo these changes once pystac handles this better.

PR #1915, opened 2022-08-04
fix: Prevent metadata duplication

Jimlinz:
Pystac `normalize_hrefs` resolves all links and mutates the entire tree, causing `<item>.json` and `collection.json` for all datasets within the catalog to update. `save_object` only saves the particular entity within the catalog, preventing unnecessary metadata duplication. 

## Reference
https://github.com/stac-utils/pystac/issues/284
https://github.com/stac-utils/pystac/pull/294
https://github.com/stac-utils/pystac/issues/90


geostore/populate_catalog/task.py
```diff
@@ -95,8 +95,4 @@ def handle_message(metadata_key: str) -> None:
     if root_catalog.get_child(dataset_metadata.id) is None:
         root_catalog.add_child(child=dataset_metadata, strategy=GeostoreSTACLayoutStrategy())
 
-    root_catalog.normalize_hrefs(
```
billgeo, line 98: We do want to normalize hrefs across the Geostore (hopefully we have a test to enforce that). Have you checked that we have code and/or a test for that?

billgeo: We definitely don't need to normalize every href every time we add a dataset though, just need to do the catalog.json and all the STAC file in the child that is being added. Can ignore all the siblings.

Jimlinz:
Changing PR to _draft_ since tests are failing. `save_object` appears to be a [suggestion](https://github.com/stac-utils/pystac/issues/90#issuecomment-614673898) to work around #1818; however, this is causing tests to fail. More work is needed to figure out why the tests are failing and if there is a better solution for the problem above.

geostore/populate_catalog/task.py
Jimlinz: This was addressed in https://github.com/stac-utils/pystac/pull/294 (if I read the PR correctly). Nevertheless, tests are failing which means more work is needed to figure out what is needed to fix this issue.

l0b0:
Looking through the test failures:

- `should_update_root_catalog_with_new_version_catalog [dataset version links]` fails because an [optional `title`](https://github.com/radiantearth/stac-spec/blob/789e90fba4c7119bc925d26f75f47bacf44c91e6/catalog-spec/catalog-spec.md#link-object) is missing. Possible solutions:
   - Ignore the irrelevant field in the test. This is probably the simplest, but means the test might have to change again if PySTAC changes.
   - Change the test to only check the URLs rather than the whole `link` array. More robust, but also more work and less coverage.
- `should_update_root_catalog_with_new_version_collection [item links]` fails for two separate reasons:
   - Missing `title`, as above.
   - **Wrong root link.** This is an actual issue, and needs fixing.
- `should_successfully_run_dataset_version_creation_process_and_again_with_partial_upload` fails because of a **missing/wrong root link** (`expected_link_object` is never `in (load(smart_open.open(f"{storage_bucket_prefix}{dataset_title}/{item_metadata_filename}", mode="rb")))[STAC_LINKS_KEY]`). Also needs fixing.

In summary, to make this work ASAP we could:

1. Modify tests to remove the `title` from the relevant expected `links` entry.
2. Change the code which writes metadata files (`geostore.import_metadata_file.task.importer`) to modify the root URL as it's currently specified in the tests.

tests/test_populate_catalog.py
```diff
@@ -269,15 +265,23 @@ def should_update_root_catalog_with_new_version_catalog(subtests: SubTests) -> N
             mode="rb",
         ) as updated_root_metadata_file:
             catalog_json = load(updated_root_metadata_file)
-            assert catalog_json[STAC_LINKS_KEY] == expected_root_catalog_links
+            updated_catalog_json = []
+            for i in catalog_json[STAC_LINKS_KEY]:
+                del i["title"]
```
l0b0, line 270: Rather than deleting the title from the actual state you could simply remove the non-existent title from the expected state. That should result in less code overall, and very slightly more coverage.

geostore/populate_catalog/task.py
```diff
@@ -95,8 +95,4 @@ def handle_message(metadata_key: str) -> None:
...
-        f"{S3_URL_PREFIX}{Resource.STORAGE_BUCKET_NAME.resource_name}",
-        strategy=GeostoreSTACLayoutStrategy(),
-    )
```
l0b0, line 101: Isn't there some missing code to actually replace the root `href`?

```
>               assert item_json[STAC_LINKS_KEY] == expected_item_links
E               AssertionError: assert [{'href': './CCRrWAn9FPPOyNFd1Q6I.json',\n  'rel': 'root',\n  'type': 'application/json'},\n {'href': './CCRrWAn9FPPOyNFd1Q6I.json',\n  'rel': 'parent',\n  'type': 'application/json'}] == [{'href': '../catalog.json', 'rel': 'root', 'type': 'application/json'},\n {'href': './CCRrWAn9FPPOyNFd1Q6I.json',\n  'rel': 'parent',\n  'type': 'application/json'}]
E                 At index 0 diff: {'rel': 'root', 'href': './CCRrWAn9FPPOyNFd1Q6I.json', 'type': 'application/json'} != {'rel': 'root', 'href': '../catalog.json', 'type': 'application/json'}
E                 Full diff:
E                   [
E                 -  {'href': '../catalog.json',
E                 +  {'href': './CCRrWAn9FPPOyNFd1Q6I.json',
E                     'rel': 'root',
E                     'type': 'application/json'},
E                    {'href': './CCRrWAn9FPPOyNFd1Q6I.json',
E                     'rel': 'parent',
E                     'type': 'application/json'},
E                   ]
```

tests/test_populate_catalog.py
Jimlinz: Do you mean something like this https://github.com/linz/geostore/commit/bb4fd6295f1b8d88c724874e82bd5b198bf2d37e?
It was my initial attempt, but it didn't work. Maybe I have missed something? 

`del` didn't work, had to change it to `pop`

l0b0: Something like bb4fd62, except some of the titles should still be there.

geostore/populate_catalog/task.py
```diff
@@ -63,7 +63,7 @@ def get_collection_href(self, col: Collection, parent_dir: str, is_root: bool) -
         assert not is_root
         return str(col.get_self_href())
 
-    def get_item_href(self, item: Item, parent_dir: str) -> str:
+    def get_item_href(self, item: Item, parent_dir: str) -> str:  # pragma: no cover
```
Jimlinz, line 66: `GeostoreSTACStrategy` `get_item_href` is not called after `normalized_href` is removed; however, this looks like it may be relevant again in the future as the codebase grows (and keeping this for completeness). Excluding this function from coverage, for now.

geostore/import_metadata_file/task.py
```diff
@@ -16,6 +16,7 @@
 from ..types import JsonObject
 
 S3_BODY_KEY = "Body"
+CATALOG_FILENAME = "catalog.json"
```
l0b0, line 19: This constant already exists in `geostore/populate_catalog/task.py`, so it should probably be moved to a file in the `geostore` directory and imported.

geostore/populate_catalog/task.py
l0b0: I would not leave dead code in place for "completeness". We'll have the repo history, so it can be recovered if we ever need it. Most likely we never will.

tests/test_processing_stack.py
```diff
@@ -483,10 +511,9 @@ def should_successfully_run_dataset_version_creation_process_and_again_with_part
 
             #  Wait for Item metadata file to include a link to geostore root catalog
             expected_link_object = {
-                STAC_HREF_KEY: f"../{CATALOG_FILENAME}",
+                STAC_HREF_KEY: CATALOG_FILENAME,
```
l0b0, line 514: I don't think this is right. The root should always be the actual root catalogue file, which is in the parent directory of the dataset.

geostore/populate_catalog/task.py
Jimlinz: It looks like `GeostoreSTACLayoutStrategy` overrides `HrefLayoutStrategy`, and can only be instantiated if it overrides all the abstract methods in the parent class??

geostore/import_metadata_file/task.py
Jimlinz: Have changed this to reference `geostore/populate_catalog/task.py` (for now). Good idea to move this to another file, but a quick look shows it is reference in quite a few places, so I might move this change to another PR.

tests/test_processing_stack.py
Jimlinz: You are correct about this. Thanks for flagging it. The problem was caused by calling `update_root_link` before another function which `change_href_to_basename`. Took some time to figure out, but makes more sense now.

geostore/import_metadata_file/task.py
Jimlinz: Spoke too soon. The value needs to be accessible via the lambda function. Changing this back to `../catalog.json` for now and will create another PR to address the above.

geostore/populate_catalog/task.py
Jimlinz: Took a while for me to trace this down. There are two options to tackle this bug: 
`root_catalog.save_object()` or `root_catalog.save(catalog_type=CatalogType.SELF_CONTAINED)` see `populate_catalog/task.py`

The first option `save_object` is the workaround suggested by Pystac (see https://github.com/stac-utils/pystac/issues/284). This would be the ideal implementation as none of the metadata gets duplicated during dataset version creation (at least in in my own tests); however, from what I can tell, this options eludes `importer` from being called in `import_metadata_file/task.py` (which is why you are seeing the error message above). Why this may be the case exactly, I am still slightly unclear. I have tried tracing this down but with no success thus far. Perhaps you might have a better understanding of how Geostore utilizes Pystac to shine some light on why this may be? I will continue to explore and will update if I am able to work this out.

The second option `root_catalog.save(catalog_type=CatalogType.SELF_CONTAINED)` is the option I went with in this PR. This appears to only be a partial fix. It prevents `item.json` from being updated with every new dataset version creation; however, it doesn't prevent `collection.json` from being updated every time a new dataset is added or updated.

geostore/populate_catalog/task.py
l0b0: In that case, how about replacing the method body with `raise NotImplementedError`? That way we'll at least know to remove the coverage pragma once it's used again.

l0b0 (dismissed):

billgeo (approved):

AmrouEmad (approved):

PR #1992, opened 2022-09-06
feat: Use lock file for Pip installs

l0b0:
## Issues

Closes https://github.com/linz/geostore/issues/1991

## Reference


Jimlinz:
Chicken and egg problem?

l0b0:
> Chicken and egg problem?

Heh, not really. The only issue was that I had to fall back to the old install method after checking out the current production commit, which was easy.

.github/workflows/prod-upgrade-deploy-test.yml
```diff
@@ -86,9 +86,10 @@ jobs:
             ${{ runner.os }}-pip-${{ env.PROD_PYTHON_VERSION }}-${{ secrets.CACHE_SEED }}-
 
       - name: Install Python dependencies
+        # TODO: Remove `|| […]` once deployed to production
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install poetry
+          python -m pip install --requirement=geostore/pip.txt || python -m pip install --upgrade pip
+          python -m pip install --requirement=geostore/poetry.txt || python -m pip install poetry
```
Jimlinz, line 92: Just wondering why `||` is needed here. Is there an expectation that one might fail and we fallback to the other method of installing pip and poetry? Would we expect any discrepancy between using the first installation method over the other? Mostly asking out of curiosity....

l0b0: There's a hint in the comment above:

> TODO: Remove `|| […]` once deployed to production

It's explained further in the commit message:

> fix: Fall back to old install method on old commit
>
> After checking out the current production commit we don't have the `.txt` lock files available.

Jimlinz (approved):

PR #2023, opened 2022-09-14
fix: Dataset version should allow catalog collection only

AmrouEmad:
## Reference


geostore/check.py
```diff
@@ -13,3 +13,5 @@ class Check(Enum):
     STAGING_ACCESS = "staging bucket access"
     UNKNOWN_CLIENT_ERROR = "unknown client error"
     UNKNOWN_MULTIHASH_ERROR = "unknown multihash error"
+    UPLOADED_ASSETS_SHOULD_BE_CATALOG_OR_COLLECTION_ERROR = "assets should be catalog or collection"
```
l0b0, line 16: An *asset* should never be a catalogue or collection - an asset is a non-metadata file like a JPEG or SVG. How about something like `INVALID_STAC_ROOT_TYPE`? The error message in `geostore/check_stac_metadata/utils.py` already expands on what that means.

The same issue applies to `InvalidAssetFileError`.

geostore/check_stac_metadata/utils.py
```diff
@@ -69,6 +67,14 @@ def maybe_convert_relative_url_to_absolute(url_or_path: str, parent_url: str) ->
     return f"{dirname(parent_url)}/{url_or_path}"
 
 
+def is_url_start_with_s3(metadata_url: str) -> bool:
```
l0b0, line 70: How about `is_s3_url`?

l0b0 (commented):
Good stuff, just have some naming issues. You'll need to change the PR title as well, to fit the conventional commit format.

geostore/check_stac_metadata/utils.py
```diff
@@ -138,12 +143,34 @@ def run(self, metadata_url: str) -> None:
...
-                    "error": NO_ASSETS_FOUND_ERROR_MESSAGE,
+                    "error": Check.NO_ASSETS_FOUND_ERROR_MESSAGE,
                     GIT_COMMIT: get_param(ParameterName.GIT_COMMIT),
                 },
             )
             return
 
+        s3_response = self.get_object(metadata_url)
+        stac_type = self.get_s3_url_as_object_json(metadata_url, s3_response)[STAC_TYPE_KEY]
+        if not is_instance_of_catalog_or_collection(stac_type):
```
l0b0, line 154: There's no test of this branch.

geostore/check_stac_metadata/utils.py
AmrouEmad: Updated. Thanks

geostore/check.py
AmrouEmad: Couldn't agree more. Updated.

geostore/check.py
```diff
@@ -13,3 +13,5 @@ class Check(Enum):
...
+    NO_ASSETS_FOUND_ERROR_MESSAGE = "No assets found in dataset"
```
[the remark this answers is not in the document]
AmrouEmad: Updated. Thanks.

geostore/check.py
```diff
@@ -13,3 +13,5 @@ class Check(Enum):
     STAGING_ACCESS = "staging bucket access"
     UNKNOWN_CLIENT_ERROR = "unknown client error"
     UNKNOWN_MULTIHASH_ERROR = "unknown multihash error"
+    INVALID_STAC_ROOT_TYPE = "Root type should be catalog or collection"
```
l0b0, line 16: The value should be formatted like the other ones, so it shouldn't be capitalised. By the way, can you please re-sort the enums so they're still alphabetical?

geostore/check_stac_metadata/utils.py
AmrouEmad: Updated. Thanks

geostore/check_stac_metadata/utils.py
```diff
@@ -69,6 +67,18 @@ def maybe_convert_relative_url_to_absolute(url_or_path: str, parent_url: str) ->
     return f"{dirname(parent_url)}/{url_or_path}"
 
 
+def is_s3_url(metadata_url: str) -> bool:
+    return metadata_url[:5] == S3_URL_PREFIX
+
+
+def is_empty(asset_list: List[Dict[str, str]]) -> int:
+    return len(asset_list)
```
l0b0, line 75: I don't understand this. Why is this called `is_empty`? And why is `len(asset_list)` pulled out as a function?

geostore/check.py
l0b0: The list isn't quite sorted yet; `NO_ASSETS_IN_DATASET` should be in the middle. (This is why I recommend leaving the original commenter to resolve conversations - it ensures that each comment is verified as addressed by the commenter.)

geostore/check_stac_metadata/utils.py
```diff
@@ -138,15 +147,41 @@ def run(self, metadata_url: str) -> None:
...
+                    "error": Check.NO_ASSETS_IN_DATASET,
                     GIT_COMMIT: get_param(ParameterName.GIT_COMMIT),
                 },
             )
             return
 
+        stac_type = self.get_stac_type_by_url(metadata_url)
+        if not is_instance_of_catalog_or_collection(stac_type):
+            error_message = (
+                f"Uploaded Assets should be catalog.json or collection.json”: “{metadata_url}”"
```
l0b0, line 159: The first "”" probably shouldn't be there.

geostore/check_stac_metadata/utils.py
```diff
@@ -251,6 +277,19 @@ def validate(self, url: str) -> None:  # pylint: disable=too-complex
...
+            object_json: JsonObject = load(
+                s3_response.response,
+                object_pairs_hook=self.duplicate_object_names_report_builder(url),
+            )
+        except JSONDecodeError as error:
+            self.validation_result_factory.save(
+                url, Check.JSON_PARSE, ValidationResult.FAILED, details={MESSAGE_KEY: str(error)}
+            )
+            raise
+        return object_json
```
l0b0, line 291: The `return` can just be in the `try` block - it doesn't change the semantics of the code.

tests/test_check_stac_metadata_utils_logging.py
```diff
@@ -101,9 +103,10 @@ def should_log_non_s3_url_prefix_validation() -> None:
     with patch("geostore.check_stac_metadata.utils.LOGGER.error") as logger_mock, patch(
         "geostore.check_stac_metadata.utils.processing_assets_model_with_meta"
     ):
-        STACDatasetValidator(
-            hash_key, url_reader, MockAssetGarbageCollector(), MockValidationResultFactory()
-        ).run(metadata_url)
+        with raises(InvalidSTACRootTypeError):
```
l0b0, line 106: This `with` could just be part of the parent `with`, as in `should_raise_invalid_stack_root_type_error_for_non_collection_or_catalog`.

l0b0:
Also, the PR title still needs to change.

geostore/check_stac_metadata/utils.py
AmrouEmad: You are right, but to pull functions out is more easier when mocking it.

geostore/check_stac_metadata/utils.py
AmrouEmad: Thanks for spotting that. Updated

tests/test_check_stac_metadata_utils_logging.py
AmrouEmad: Updated. Thanks.

geostore/check.py
AmrouEmad: Updated. Thanks.

geostore/check_stac_metadata/utils.py
AmrouEmad: Updated. Thanks.

PR #2029, opened 2022-09-15
feat: Compare s3 object etag in S3StacIO

Jimlinz:
This PR prevents PyStac metadata from being re-written in S3 if one already exists and is identical. A condition is added to `S3StacIO` to calculate check the `etag` of the file before writing it in S3 (if object already exists in the bucket).

Note: This only works when `save()` is called on the catalog. It doesn't prevent other identical files (metadata or otherwise) from being rewritten in S3 via other method (e.g. when user runs datasets-version on an existing data)

Closes https://github.com/linz/geostore/issues/1995

## Reference

https://stackoverflow.com/questions/12186993/what-is-the-algorithm-to-compute-the-amazon-s3-etag-for-a-file-larger-than-5gb
https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-chunksize


tests/test_s3_utils.py
```diff
@@ -167,3 +173,120 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+    ):
+        s3_object = check_if_s3_object_exists(
+            Resource.STAGING_BUCKET_NAME.resource_name,
+            f"{key_prefix}/{collection_metadata_filename}",
+        )
+
+        assert s3_object is True
+
+
+def should_return_accurate_etag_value() -> None:
```
l0b0, line 200: This should probably be several tests with one assertion each.

tests/test_s3_utils.py
```diff
@@ -167,3 +173,120 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+    # parse custom chunk_size to mimic big file behaviour in test
+    etag_from_big_file = calculate_s3_etag(byte_representation_in_big_file, chunk_size=8)
+
+    assert etag_from_empty_file == expected_empty_file_etag
+    assert etag_from_small_file == expected_small_file_etag
+    assert etag_from_big_file == expected_big_file_etag
+
+
+@mark.infrastructure
+def should_return_etag_if_s3_object_exists_or_return_empty_string() -> None:
```
l0b0, line 221: Isn't this two unrelated tests in one?

tests/test_s3_utils.py
```diff
@@ -167,3 +173,120 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+        # Write object to bucket - v1
+        S3StacIO().write_text(dest=s3_obj_url, txt=sample_obj_str)
+
+        s3_etag_one = get_s3_etag(
+            Resource.STAGING_BUCKET_NAME.resource_name,
+            collection_metadata_filename,
+            get_log(),
+        )
+
+        assert s3_etag_one == "2d0503d7761d6240b7a6bffca8ba25fa"
```
l0b0, line 274: Why do this assertion?

tests/test_s3_utils.py
```diff
@@ -167,3 +173,120 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+
+        s3_response = s3_client.head_object(Bucket=bucket, Key=collection_metadata_filename)
+        obj_modified_one = s3_response["LastModified"]
+
+        # Write same object to bucket again - v2
+        S3StacIO().write_text(dest=s3_obj_url, txt=sample_obj_str)
```
l0b0, line 280: We should be able to reuse the `S3StacIO` object across writes.

tests/test_s3_utils.py
```diff
@@ -167,3 +173,120 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+        s3_response = s3_client.head_object(Bucket=bucket, Key=collection_metadata_filename)
+        obj_modified_two = s3_response["LastModified"]
+
+        # Ensure original object is not rewritten
+        assert obj_modified_one == obj_modified_two
```
l0b0, line 285: Would checking the length of the `list_object_versions` response be a more direct test of the use case?

geostore/s3_utils.py
```diff
@@ -52,3 +54,58 @@ def s3_url_reader(staging_url: str) -> GeostoreS3Response:
...
+
+
+def calculate_s3_etag(
+    body: bytes,
+    # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-chunksize
+    chunk_size: int = 8 * 1024 * 1024,  # Default value
+) -> str:
+
+    md5s = []
+    data = [body[i : i + chunk_size] for i in range(0, len(body), chunk_size)]
```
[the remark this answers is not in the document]
Jimlinz: This is a good point. Is the concern here mainly around large files? Or is there any other reason to avoid doing so?
(i.e. if user submit files over 1gb, all of it would be written to memory, exceeding lambda limit?)

geostore/s3_utils.py
```diff
@@ -52,3 +54,58 @@ def s3_url_reader(staging_url: str) -> GeostoreS3Response:
...
+        )
+        # rather than raise, we return an empty string, indicating that the etag is different
+        # thus allowing the next step to continue rather than stalling the entire process
+        return ""
+
+
+def calculate_s3_etag(
+    body: bytes,
+    # https://docs.aws.amazon.com/cli/latest/topic/s3-config.html#multipart-chunksize
+    chunk_size: int = 8 * 1024 * 1024,  # Default value
```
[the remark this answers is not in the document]
Jimlinz: >One way to make sure we're actually implementing an ETag calculation which matches AWS would be to have tests which upload empty, single-chunk and multi-chunk files to AWS...

This thought did occurred to me as I was working on this. I wasn't too sure about having (non trivial size) tests files either stored in a bucket somewhere or added to the git repo. Generating random file content is much more appealing, but I was under the impression that `md5` is used as part of calculating the eTag? Perhaps I have misunderstood something here??

I ended up writing the test using a byte representation of files (see `test_s3_utils.py::should_return_accurate_etag_value`), which appears to be enough for me to calculate the corresponding `etag` for testing purposes. Would you recommend I revisit this with the suggestions above? (i.e. either import asset file for testing or generate random file content)

>Also, should this really be configurable?

I saw a reference somewhere that suggests different clients use different default values (e.g. using aws cli vs third party software). Unfortunately I have lost the link that suggest this and a quick search comes up blank. That said, I am aware that this is unlikely to affect _geostore_ since s3 to s3 transfer is the only currently supported transfer method. I have left this configurable with default value mostly to allow the test to simulate a multi chunk file without having to upload or generate files larger than 8mb.

Perhaps a discussion might be helpful here?

tests/test_s3_utils.py
Jimlinz: Have opted for `VersionId` instead. It seems they are different even if the same object is uploaded again. `list_object_versions` could potentially return a fairly sizeable list, depending how many times the metadata is (re)written over in s3.

geostore/s3_utils.py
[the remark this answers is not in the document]
l0b0: Failing the job because of big files would definitely be an issue. Fortunately that's one we could work around (at least in the short term) by bumping the max memory and paying more for the Lambdas. Stream processing is also usually faster than bulk file processing, so there's also a performance concern.

geostore/s3_utils.py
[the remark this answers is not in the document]
l0b0: > > One way to make sure we're actually implementing an ETag calculation which matches AWS would be to have tests which upload empty, single-chunk and multi-chunk files to AWS...
> 
> This thought did occurred to me as I was working on this. I wasn't too sure about having (non trivial size) tests files either stored in a bucket somewhere or added to the git repo. Generating random file content is much more appealing, but I was under the impression that `md5` is used as part of calculating the eTag? Perhaps I have misunderstood something here??

We wouldn't create any files in the filesystem - simply push a stream of `os.urandom` to an S3 bucket as part of the test. We don't really care what the contents are (except for the empty file test), just that it uses our upload functionality and that the ETag we calculate is the same as reported by AWS.

> I ended up writing the test using a byte representation of files (see `test_s3_utils.py::should_return_accurate_etag_value`), which appears to be enough for me to calculate the corresponding `etag` for testing purposes. Would you recommend I revisit this with the suggestions above? (i.e. either import asset file for testing or generate random file content)

Yes. Basically, there's no easy way for someone reading the tests to say "yes, this test does what it says it should do", because the inputs and expected outputs have no connection to each other which can be easily understood. If we instead have a test which just sends some random data to S3 and then just asserts that our ETag calculation gives the same result as S3, then we've proven just that: our implementation is (at least for the values we're testing) equivalent to the AWS implementation. Which I think is the best we can hope for, since if we ever start using a different chunk size or some other way to upload things, we'll end up with a different ETag calculation.

(Ideally we'd simply call a library function which implements ETag calculation, but I guess you've not been able to find a good/official one? I could only find third party ones. :shrug:)

> > Also, should this really be configurable?
> 
> I saw a reference somewhere that suggests different clients use different default values (e.g. using aws cli vs third party software). Unfortunately I have lost the link that suggest this and a quick search comes up blank. That said, I am aware that this is unlikely to affect _geostore_ since s3 to s3 transfer is the only currently supported transfer method. I have left this configurable with default value mostly to allow the test to simulate a multi chunk file without having to upload or generate files larger than 8mb.

I don't think such a test is very useful, because it's not testing the core of the algorithm - that we get the same result from our implementation as from S3 itself.

geostore/s3_utils.py
```diff
@@ -52,3 +54,58 @@ def s3_url_reader(staging_url: str) -> GeostoreS3Response:
     staging_s3_client = get_s3_client_for_role(s3_role_arn)
     geostore_s3_client = get_s3_client_for_role(get_param(ParameterName.S3_USERS_ROLE_ARN))
     return s3_url_reader
+
+
+def check_if_s3_object_exists(bucket: str, key: str) -> bool:
```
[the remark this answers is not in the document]
Jimlinz: Good point. Have removed.

tests/test_s3_utils.py
```diff
@@ -167,3 +175,241 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+            )
+
+            s3_etag = get_s3_etag(
+                s3_bucket=storage_bucket, s3_object_key=f"{asset_filename}", logger=get_log()
+            )
+
+            # ensure single chunk size file
+            assert (
+                len(local_etag) == 34
+            )  # single chunk etag is 32 char, (+2) with quotes as part of etag
```
l0b0, line 252: It took me a while to remember this, so it might be good to mention in the comment that this length is specifically only for single-chunk files. You could even avoid the comments by pulling out `34` as something like `SINGLE_CHUNK_ETAG_LENGTH`. and `36` as `MULTI_CHUNK_ETAG_LENGTH`.

tests/test_s3_utils.py
```diff
@@ -167,3 +175,241 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+            )  # empty file etag should always be the same
+            assert local_etag == s3_etag
+        finally:
+            delete_s3_key(Resource.STORAGE_BUCKET_NAME.resource_name, asset_filename, s3_client)
+
+
+@mark.infrastructure
+def should_return_accurate_etag_value_for_single_chunk_size_file(s3_client: S3Client) -> None:
+    asset_filename = any_safe_filename()
+    asset_contents = any_file_contents(byte_count=1000000)
```
l0b0, line 223: It's usually good to change "magic" numbers like this to refer to an existing variable or constant, to clarify *why* the value was chosen. In this case it would be cool if the size of this file was *exactly* one chunk (you'd have to pull out `8_388_608` as a constant, maybe something like `S3_DEFAULT_CHUNK_SIZE`?), and the size of the file in the other test was just `S3_DEFAULT_CHUNK_SIZE + 1`. That way we can guarantee that if the chunk size changes *at all* one of the tests would start failing.

geostore/s3_utils.py
```diff
@@ -52,3 +53,43 @@ def s3_url_reader(staging_url: str) -> GeostoreS3Response:
...
+        return s3_response["ETag"]
+    except ClientError as error:
+        if error.response["Error"]["Code"] != "404":
+            logger.debug(
+                f"Unable to fetch eTag for “{s3_object_key}” in s3://{s3_bucket} due to “{error}”",
+                extra={GIT_COMMIT: get_param(ParameterName.GIT_COMMIT)},
+            )
+        # rather than raise, we return an empty string, indicating that the etag is different
+        # thus allowing the next step to continue rather than stalling the entire process
+        return ""
```
l0b0, line 72: Usually we'd return `None` for an unknown value; meaning the function return value type would be `Optional[str]`. Then you won't have to do length tests to check whether you got an actual ETag back, you'd just do `s3_etag is not None`. Does that make sense?

tests/test_s3_utils.py
Jimlinz: Turns out at exactly single chunk (i.e. `8_388_608`), the etag length is `36` (i.e. `34charetag-1`); where the -1 signify the number of chunk. This makes sense. I have adjusted the test to test for file smaller than single chunk (i.e. `S3_DEFAULT_CHUNK_SIZE - 1`)

geostore/pystac_io_methods.py
```diff
@@ -33,4 +41,11 @@ def write_text(  # type: ignore[override]
     ) -> None:
         url = dest.href if isinstance(dest, Link) else dest
         bucket, key = get_bucket_and_key_from_url(url)
+        if check_if_s3_object_exists(bucket, key):
+            s3_etag = get_s3_etag(bucket, key, LOGGER)
+            local_etag = calculate_s3_etag(txt.encode())
+
+            if s3_etag == local_etag:
+                return
```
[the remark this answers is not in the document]
Jimlinz: Have changed the logic here slightly. First, check if object exists in s3 bucket. This should either return the `etag` of the object or `None` (in the case of `404`). Any other error messages (e.g. 403 permission denied or 500 server error) will get logged (see `get_s3_etag`).

tests/test_s3_utils.py
Jimlinz: Have removed. No longer relevant since the tests are now calculating etag based on randomly generated files.

tests/test_s3_utils.py
```diff
@@ -167,3 +181,240 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+        f"Unable to fetch eTag for “{collection_metadata_filename}” "
+        f"in s3://{bucket} due to “{error}”"
+    )
+
+    logger_mock = MagicMock()
+
+    get_s3_etag(s3_bucket=bucket, s3_object_key=collection_metadata_filename, logger=logger_mock)
+
+    # Then
+    with subtests.test(msg="log"):
```
l0b0, line 417: You shouldn't need a `subtests.test` for a single assertion like this.

tests/test_s3_utils.py
```diff
@@ -167,3 +181,240 @@ def should_raise_any_client_error_other_than_no_such_key(
...
         s3_url_reader = get_s3_url_reader(get_s3_role_arn(), any_dataset_title(), logger_mock)
         s3_url_reader(any_s3_url())
+
+
+@mark.infrastructure
+def should_return_accurate_etag_value_for_empty_file(s3_client: S3Client) -> None:
+    asset_filename = any_safe_filename()
+    asset_contents = any_file_contents(byte_count=0)
+    expected_empty_file_etag = (
+        '"d41d8cd98f00b204e9800998ecf8427e"'  # we know the etag of empty file
```
l0b0, line 191: "empty file" should probably be "an empty file" or "empty files".

tests/test_s3_utils.py
```diff
@@ -45,6 +55,12 @@
 else:
     ClientErrorResponseError = ClientErrorResponseTypeDef = dict
 
+SINGLE_CHUNK_ETAG_LENGTH = 34  # single chunk etag is "32_char", (+2) with quotes as part of etag
```
l0b0, line 58: What is "32_char"? If you mean "32 characters", it's probably easier to say so. You might also refer to a definition of ETag for both constants, to provide more info if needed.

geostore/s3_utils.py
```diff
@@ -52,3 +53,43 @@ def s3_url_reader(staging_url: str) -> GeostoreS3Response:
...
+        # thus allowing the next step to continue rather than stalling the entire process
+        return None
+
+
+def calculate_s3_etag(body: bytes) -> str:
+    # https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#multipart-chunksize
+    s3_default_chunk_size = 8_388_608  # Default value is 8 * 1024 * 1024
+
+    if body == b"":
+        return f'"{hashlib.md5().hexdigest()}"'
```
l0b0, line 80: Since this value is static and equal to the value in `should_return_accurate_etag_value_for_empty_file` you could just pull out a constant for it. In that case the `assert local_etag == expected_empty_file_etag` might be redundant. What do you think?

tests/test_s3_utils.py
l0b0: WTF. This might be a bug in how S3 calculates ETags, but I've added another comment for it.

tests/test_s3_utils.py
```diff
@@ -167,3 +181,240 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+            delete_s3_key(Resource.STORAGE_BUCKET_NAME.resource_name, asset_filename, s3_client)
+
+
+@mark.infrastructure
+def should_return_accurate_etag_value_for_file_smaller_than_single_chunk_size(
+    s3_client: S3Client,
+) -> None:
+    asset_filename = any_safe_filename()
+    # file size exactly at single chunk returns 34 char etag rather than 32 char (i.e. 32CharEtag-1)
+    # where -1 at the end of the etag denotes 1 chunk
```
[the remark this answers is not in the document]
Jimlinz: At exactly `s3_default_chunk_size = 8_388_608` things become weird. The 32 character ETag is calculated as if the file is smaller than one chunk, but a `-1` suffix is added to the ETag to indicate the file is 1 chunk. I have edited `s3_utils.py` to reflect this... hopefully it makes sense and not adding to the confusion. Weird special case indeed.

Added `should_return_accurate_etag_value_for_single_chunk_size_file` to `test_s3_utils.py`

tests/test_s3_utils.py
```diff
@@ -167,3 +181,240 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+                new_key=asset_filename,
+                source_s3_client=get_s3_client_for_role(get_s3_role_arn()),
+            )
+
+            s3_etag = get_s3_etag(
+                s3_bucket=storage_bucket, s3_object_key=f"{asset_filename}", logger=get_log()
+            )
+
+            # ensure multi chunk size file
+            assert len(local_etag) >= MULTI_CHUNK_ETAG_LENGTH
```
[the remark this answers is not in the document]
Jimlinz: The original intent is to cater for multi chunks (potentially more than 10 chunks), but have realised that this test is indeed testing a 2-chunk file, so yes `==` is accurate and also easier for others to read. Have updated.

tests/test_s3_utils.py
```diff
@@ -167,3 +181,240 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+
+    s3_obj_url = f"s3://{bucket}/{collection_metadata_filename}"
+    asset_contents = any_file_contents()
+
+    try:
+        # Write object to bucket - v1
+        s3_stac_io.write_text(dest=s3_obj_url, txt=str(asset_contents))
+
+        s3_response = s3_client.get_object(Bucket=bucket, Key=collection_metadata_filename)
+        obj_version_id_one = s3_response["VersionId"]
```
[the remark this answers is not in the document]
Jimlinz: Have opted for `first_write_attempt` and `second_write_attempt` instead, which may be a bit more accurate - since we are essentially making two write attempts (and second one should not happen if object already exists and etag the same). But yes, "one" and "two" is ambiguous..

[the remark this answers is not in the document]
l0b0: The word "version" should probably be in the name though - the current name doesn't describe the contents of the variable IMO. Maybe something (very verbose, maybe possible to shorten) like `version_id_after_nth_write_attempt`?

tests/test_s3_utils.py
```diff
@@ -45,6 +55,15 @@
 else:
     ClientErrorResponseError = ClientErrorResponseTypeDef = dict
 
+# https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html
+# single chunk etag is 32 characters in length, (+2) with quotes as they are part of etag
+SINGLE_CHUNK_ETAG_LENGTH = 34
+# multi chunk etag includes a -x suffix (+2 characters), where x specifies the number of chunks
+MULTI_CHUNK_ETAG_LENGTH = 36  # assuming less than 10 chunks
```
l0b0, line 62: How about a "minimum" word in the constant to clarify this?

geostore/s3_utils.py
```diff
@@ -52,3 +53,48 @@ def s3_url_reader(staging_url: str) -> GeostoreS3Response:
...
+        # thus allowing the next step to continue rather than stalling the entire process
+        return None
+
+
+def calculate_s3_etag(body: bytes) -> str:
+    # https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#multipart-chunksize
+    s3_default_chunk_size = 8_388_608  # Default value is 8 * 1024 * 1024
+
+    if body == b"":
+        return '"d41d8cd98f00b204e9800998ecf8427e"'  # we know the etag of empty file
```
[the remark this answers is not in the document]
Jimlinz: Have updated. Hopefully I haven't misunderstood the comment.

l0b0 (approved):
Excellent stuff!

geostore/s3_utils.py
```diff
@@ -52,3 +55,48 @@ def s3_url_reader(staging_url: str) -> GeostoreS3Response:
...
+                extra={GIT_COMMIT: get_param(ParameterName.GIT_COMMIT)},
+            )
+        # rather than raise, we return an empty string, indicating that the etag is different
+        # thus allowing the next step to continue rather than stalling the entire process
+        return None
+
+
+def calculate_s3_etag(body: bytes) -> str:
+    # https://awscli.amazonaws.com/v2/documentation/api/latest/topic/s3-config.html#multipart-chunksize
+    s3_default_chunk_size = 8_388_608  # Default value is 8 * 1024 * 1024
```
l0b0, line 79: This is a dupe of `S3_DEFAULT_CHUNK_SIZE`

tests/test_s3_utils.py
```diff
@@ -167,3 +189,248 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+                new_key=asset_filename,
+                source_s3_client=get_s3_client_for_role(get_s3_role_arn()),
+            )
+
+            s3_etag = get_s3_etag(
+                s3_bucket=storage_bucket, s3_object_key=f"{asset_filename}", logger=get_log()
+            )
+
+            # file size exactly at single chunk returns 34 characters etag rather than 32 characters
+            # (i.e. "656dadd6d61e0ebfd29264e34d742df3-1") where -1 suffix denotes 1 chunk
```
l0b0, line 299: We might want this at the top of the test to clarify why we have this special case test. Also, "i.e." ("id est" == "that is") should probably be "e.g." ("exempli gratia" == "for example").

tests/test_s3_utils.py
```diff
@@ -167,3 +189,248 @@ def should_raise_any_client_error_other_than_no_such_key(
...
+                new_key=asset_filename,
+                source_s3_client=get_s3_client_for_role(get_s3_role_arn()),
+            )
+
+            s3_etag = get_s3_etag(
+                s3_bucket=storage_bucket, s3_object_key=f"{asset_filename}", logger=get_log()
+            )
+
+            # ensure single chunk size file
+            assert len(local_etag) == SINGLE_CHUNK_ETAG_LENGTH
```
l0b0, line 261: As discussed elsewhere, we probably don't need this assertion.

PR #2072, opened 2022-09-26
feat: Run pytest in parallel

Jimlinz:
So that tests run as fast as possible, we want to run them in parallel.

## Reference


Jimlinz:
Removing `root catalog` (or changing test assertion so that root catalog may contain multiple parent or child item from another test) is a prerequisite for this piece of work. Currently, some of the tests assert that the `root catalog` may only contains one or two expected parent or child item. When running tests in parallel, there is a high chance that an adjacent test would write a different parent or child item to said root catalog, thus making some of the test assertion to fail.

PR #2084, opened 2022-09-29
feat: Better logging and exception handling in populate catalog

Jimlinz:
When `populate_catalog` lambda fails, messages get stuck in SQS. This prevents subsequent datasets from being added to geostore (in completion). Without alerting in place, the problem persists until someone notices it. 

This PR implements better logging and exception handling in `populate_catalog`. The logs should capture when an exception is thrown. Elasticsearch should monitor for such problem and alert the geostore team accordingly. 

## Reference


tests/test_populate_catalog.py
```diff
@@ -652,3 +655,26 @@ def should_add_link_to_root_catalog_in_series(
...
             )
+
+
+@mark.infrastructure
+@patch("geostore.populate_catalog.task.read_file")
+def should_log_error_message_when_an_exception_is_raised_trying_to_update_catalog(
+    pystac_read_file_mock: MagicMock,
+) -> None:
+
+    error = Exception()
```
l0b0, line 666: There's an `any_exception_class` function you might want to use here. It does a couple useful things:

1. Ties the mock side effect explicitly to the `raises` clause, since you have to reference the same class in both places for the test to pass.
2. `Exception` is the superclass of *all* standard error classes, so I believe `raises(Exception)` will pass no matter which exception is raised. By using a *specific* and *unique* exception we ensure that any other raised exceptions will cause the test to fail.

(We should probably change `any_exception_class` to check that the class name isn't defined yet before creating the class, just to make sure we don't accidentally override an existing exception class during the test.)

tests/test_populate_catalog.py
```diff
@@ -652,3 +655,26 @@ def should_add_link_to_root_catalog_in_series(
...
+    with patch("geostore.populate_catalog.task.LOGGER.warning") as logger_mock:
+        with raises(Exception):
+            handle_message(any_s3_url())
+
+        logger_mock.assert_any_call(
+            expected_message,
+            extra={
+                GIT_COMMIT: get_param(ParameterName.GIT_COMMIT),
+            },
+        )
```
l0b0, line 680: FYI, Black will automatically split things into multiple lines if there's a trailing comma like on the two preceding lines. If you remove the trailing comma I believe all or most of this assertion should fit on a single line.

tests/test_populate_catalog.py
Jimlinz: Excellent. Was not aware of the `any_exception_class` function. Thanks for logging #2093. Would that be a prerequisite for this PR to be approved?

tests/test_populate_catalog.py
```diff
@@ -652,3 +655,24 @@ def should_add_link_to_root_catalog_in_series(
...
+) -> None:
+    exception_class = any_exception_class()
+    error_message = any_error_message()
+    error = exception_class(error_message)
+    expected_message = f"{LOG_MESSAGE_LAMBDA_FAILURE}: Unable to populate catalog due to “{error}”"
+
+    pystac_read_file_mock.side_effect = error
+
+    with patch("geostore.populate_catalog.task.LOGGER.warning") as logger_mock:
+        with raises(exception_class):
```
l0b0, line 673: [`raises` can be part of the previous `with`](https://github.com/linz/geostore/pull/2098#discussion_r987378850).

tests/test_populate_catalog.py
l0b0: Nah, that's definitely a separate PR.

tests/test_populate_catalog.py
```diff
@@ -652,3 +655,25 @@ def should_add_link_to_root_catalog_in_series(
...
+    expected_message = f"{LOG_MESSAGE_LAMBDA_FAILURE}: Unable to populate catalog due to “{error}”"
+
+    pystac_read_file_mock.side_effect = error
+
+    with patch("geostore.populate_catalog.task.LOGGER.warning") as logger_mock, raises(
+        exception_class
+    ):
+        handle_message(any_s3_url())
+
+        logger_mock.assert_any_call(
```
l0b0, line 677: You'll need to dedent this or it'll be skipped once `handle_message` raises `exception_class`.

Jimlinz: Good catch!

l0b0 (approved):

PR #2098, opened 2022-10-04
fix: dataset version should only accept catalog.json and collection

AmrouEmad:
## Reference


tests/test_check_stac_metadata_utils_logging.py
```diff
@@ -101,9 +103,10 @@ def should_log_non_s3_url_prefix_validation() -> None:
     with patch("geostore.check_stac_metadata.utils.LOGGER.error") as logger_mock, patch(
         "geostore.check_stac_metadata.utils.processing_assets_model_with_meta"
     ):
-        STACDatasetValidator(
-            hash_key, url_reader, MockAssetGarbageCollector(), MockValidationResultFactory()
-        ).run(metadata_url)
+        with raises(InvalidSTACRootTypeError):
```
AmrouEmad, line 106: If I merged the `raises` statement up with the update with, it will not reach to the below `assest_any_call  ` condition below.

geostore/check_stac_metadata/utils.py
```diff
@@ -294,6 +313,33 @@ def report_duplicate_object_names(object_pairs: List[Tuple[str, Any]]) -> JsonOb
 
         return report_duplicate_object_names
 
+    def check_if_contains_assets(self, metadata_url: str) -> None:
```
AmrouEmad, line 316: I prefer to export this logic into a separate function to be easier to read and test.

tests/test_check_stac_metadata_utils_logging.py
l0b0: TIL: [`with` does *not* create a scope](https://stackoverflow.com/a/45100308/96588). So you can move `raises(InvalidSTACRootTypeError)` into the top `with` statement if you dedent the `logger_mock.assert_any_call`, like this:

```python
with patch(…) as logger_mock, raises(InvalidSTACRootTypeError):
    STACDatasetValidator(…).run(metadata_url)
logger_mock.assert_any_call(…)
```

geostore/check.py
```diff
@@ -6,9 +6,11 @@ class Check(Enum):
     CHECKSUM = "checksum"
     DUPLICATE_OBJECT_KEY = "duplicate asset name"
     FILE_NOT_FOUND = "file not found in staging or storage"
+    INVALID_STAC_ROOT_TYPE = "root type should be catalog or collection"
```
l0b0, line 9: "must" rather than "should" might be helpful here.

AmrouEmad: Updated, thanks.

l0b0 (approved):

PR #2121, opened 2022-10-10
feat: Allow macrons in dataset titles

Jimlinz:
This PR is to allow the use of macron characters in dataset titles.

## Reference


geostore/dataset_properties.py
```diff
@@ -1,4 +1,5 @@
 from string import ascii_letters, digits
 
-TITLE_CHARACTERS = f"{ascii_letters}{digits}_-"
+MACRONS = "āēīōūĀĒĪŌŪ"
```
l0b0, line 3: Since `MACRONS` isn't used anywhere else you could just inline it.

l0b0 (approved):

PR #2126, opened 2022-10-10
fix: Allow step function to keep the payload when exception thrown

AmrouEmad:
## Reference


infrastructure/constructs/processing.py
```diff
@@ -441,7 +441,9 @@ def __init__(
         # STATE MACHINE
         dataset_version_creation_definition = (
             check_stac_metadata_task.add_catch(
-                errors=[Errors.TASKS_FAILED], handler=validation_summary_task
+                errors=[Errors.TASKS_FAILED],
+                handler=validation_summary_task,
+                result_path="$.error-info",
```
billgeo, line 446: Not sure why this doesn't cause an error in the tests, but should this comma be here on the end of this line?

l0b0: That's allowed in Python, and encouraged in vertical lists, because it makes the diff smaller when adding or removing list items.

billgeo: Cool. 😄

l0b0 (approved):

PR #2130, opened 2022-10-11
build: Bump nixpkgs

l0b0:
## Reference


Jimlinz:
Would a short blurb in `README.md` be helpful (for newcomers) to explain when to run `bump-nixpkgs.bash`?

l0b0:
> Would a short blurb in `README.md` be helpful (for newcomers) to explain when to run `bump-nixpkgs.bash`?

Yep. About time I added some info for dealing with various updates.

README.md
```diff
@@ -213,6 +213,27 @@ arguments above.
 
 ## Development
 
+### Third party updates
+
+When Dependabot updates any Python dependencies in pip requirements files (`*.txt`), make sure to
+run `./generate-requirements-files.bash` with the relevant path to update the version of all its
+dependencies. Sometimes this will _revert_ the file to the previous state, which means that specific
```
Jimlinz, line 220: Should we be monitoring Dependabot update and running this manually each time? I assume there isn't a way to automate this?

Jimlinz (approved):

README.md
l0b0: Some thoughts about this:

- We could definitely automate this, although I suspect it's quite a bit of work to make it reliable and to hook it into the Dependabot process.
- There is probably some more elegant way to achieve the same thing. It might be worth looking at Dependabot settings and what others are doing to deal with this.
- We might move to a different solution than Dependabot, and this problem might not exist with that other solution.

PR #2152, opened 2022-10-17
docs: Remove unused definitions

l0b0:
## Reference


billgeo (commented):
We probably should still have some kind of definition of done though shouldn't we? Maybe we should make this more user friendly rather than chuck it out completely?

l0b0:
> We probably should still have some kind of definition of done though shouldn't we? Maybe we should make this more user friendly rather than chuck it out completely?

Good idea - do you have any suggestions? Right now it is just a wall of text we never really look into, and might be more process than we need in the group.

Jimlinz:
Given that tests have to past and code peer-reviewed before it can be merged (these are strictly enforced anyway), I'd say some of the criteria are a bit redundant. Perhaps we can have a quick discussion in our next retro?? to decide what we want or need? Some sort of list might be useful, but only if it is followed....

PR #2175, opened 2022-10-17
feat: Remove dependabot commits from changelog

Jimlinz:
Changelog typically outlines what has changed since the last release. We can GitHub to automatically generate release notes every time we do a release; however, this is often filled with commits created by dependabot (bump package version). This PR excludes dependabot commits from future release notes. 

## Reference
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,4 @@
+changelog:
+  exclude:
+    authors:
+      - dependabot


.github/release.yml
```diff
@@ -0,0 +1,10 @@
+# .github/release.yml
+
+changelog:
+  exclude:
+    authors:
+      - dependabot
+  categories:
+    - title: release-
+      labels:
+        - '*'
```
l0b0, line 10: I'm a bit confused by this. Doesn't this chuck every labelled PR into a category called "release-"? Why that name, and why not include unlabelled PRs?

l0b0 (approved):

.github/release.yml
Jimlinz: Turns out I am the one who is confused about this. I was thrown off by the `Required` label in the description (GitHub documentation). Have created a separate repo to test - they were indeed not needed. It is only required if I wanted to include categories in the changelog. The documentation could have been a bit clearer, or maybe the user could be a bit better (i.e. me 😅 )

.github/release.yml
```diff
@@ -0,0 +1,6 @@
+# .github/release.yml
```
l0b0, line 1: I guess this is a copy-pasta error from the docs?

l0b0 (approved):

.github/release.yml
l0b0: Very cool that you're testing in a separate repo!

PR #2246, opened 2022-11-01
fix: Checkout pull request HEAD commit instead of merge commit

Jimlinz:
By default, GitHub Action will checkout the merge commit created for the pull request rather than the actual commit used to create the pull request. There is a slim but non zero chance that this could cause some flaky tests. This PR is to get GitHub Actions to checkout pull request HEAD commit instead of merge commit.

**Note:** This only applies to `pull_request` as per [GitHub documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request). Since geostore's workflow can be triggered by different events, a conditional step is added to ensure this change only applies to `pull_request`. 

## Reference
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
https://github.com/actions/checkout/issues/261
https://github.com/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit

### DIFF
--- a/.github/workflows/prod-upgrade-deploy-test.yml
+++ b/.github/workflows/prod-upgrade-deploy-test.yml
@@ -14,7 +14,7 @@ jobs:
       id-token: write
       contents: read
     steps:
-      - name: Check out repository
+      - name: Checkout repository
         uses: actions/checkout@v3.1.0
         with:
           ref: ${{ github.event.pull_request.head.sha }}
--- a/.github/workflows/prod-upgrade-deploy-test.yml
+++ b/.github/workflows/prod-upgrade-deploy-test.yml
@@ -17,6 +17,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v3.1.0
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
           submodules: true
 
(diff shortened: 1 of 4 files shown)


.github/workflows/cdk-bootstrap.yml
```diff
@@ -13,6 +13,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v3.1.0
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
l0b0, line 16: This won't exist for `workflow_dispatch`.

.github/workflows/prod-upgrade-deploy-test.yml
```diff
@@ -14,9 +14,10 @@ jobs:
       id-token: write
       contents: read
     steps:
-      - name: Check out repository
+      - name: Checkout repository
```
l0b0, line 17: The wording should normally be "Check out" - "checkout" is just the name of the Git sub-command.

l0b0: Seems to be arbitrary in English.

l0b0 (approved):

PR #2301, opened 2022-11-14
feat: Set lambda log retention based on deployment environment

Jimlinz:
AWS keeps all Lambda logs forever by default. This makes it very difficult to search for anything useful in CloudWatch, especially in CI account. AWS only loads up to 10,000 log groups by default, and sorting by date proves to be challenging due to pagination. 

We are unlikely to need these logs > 3 months in non prod account.


infrastructure/constructs/lambda_config.py
```diff
@@ -1,6 +1,7 @@
-from aws_cdk import Duration, aws_lambda
+from aws_cdk import Duration, aws_lambda, aws_logs
 
 PYTHON_RUNTIME = aws_lambda.Runtime.PYTHON_3_9
+LOG_RETENTION = aws_logs.RetentionDays.THREE_MONTHS
```
l0b0, line 4: Three months seems a bit short - we've had more than that between releases, and in my experience it's often useful to compare the logs of an earlier release with the current release. In our case, I'd guess the stable system will eventually get to a point where dependency releases happen every few days semi-automatically, and actual feature/bug releases happen every few months (similar to the current linz-data-lake and bde-processor-deployment repos). In that case, having logs back a year or even more might be useful. What do you think, @billgeo?

Jimlinz: I am ok changing the length of time if we can come to an agreement (anything other than keeping the logs forever) 😃

Jimlinz: Another thing to consider is that we have exceeded 10,000 log groups in CI. This makes searching the most recent logs in Cloudwatch difficult (I was under the impression that logging wasn't even working). Keeping older logs could be useful, but I would much rather have easy access to recent logs. Maybe there is a way to archive older logs >3 months??? (I will investigate)...

l0b0: We need to consider the difference between logs and log groups. I don't expect there's a way to delete log *groups* automatically after a time. We should make sure we're only logging to a handful of groups in production, and then just delete all the others. Then our log retention policy could be much more lenient. If I understand correctly it should be possible to retain production log groups without giving them a static name, since most modifications to our stack shouldn't necessitate creating new log groups.

billgeo: Sure. I would say six months would do the trick if we want to keep them between releases.

Jimlinz: What about setting log retention to 1 year for prod and 3 months for non-prod?

billgeo: Sounds fine to me.

infrastructure/constructs/lambda_config.py
```diff
@@ -1,6 +1,15 @@
...
+from geostore.environment import is_production
 
 PYTHON_RUNTIME = aws_lambda.Runtime.PYTHON_3_9
 
 DEFAULT_LAMBDA_MAX_MEMORY_MEBIBYTES = 1024
 DEFAULT_LAMBDA_TIMEOUT = Duration.seconds(60)
+
+
+def log_retention_environment() -> aws_logs.RetentionDays:
+    if is_production():
```
l0b0, line 12: It shouldn't be possible for `is_production` to change return value during a deployment, making its result effectively constant at runtime. So we could pull this out as a constant:

```python
if is_production():
    RETENTION_DAYS = aws_logs.RetentionDays.ONE_YEAR
else:
    RETENTION_DAYS = aws_logs.RetentionDays.THREE_MONTHS
```

l0b0 (approved):

PR #2334, opened 2022-11-21
feat: Destroy aws lambda log group (non-prod) on teardown

Jimlinz:
To ensure Cloudwatch logs are searchable, we want to ensure a minimal amount of log groups are kept (especially in CI). This PR sets aws lambda log group policy to destroy on stack teardown.

### DIFF
--- a/infrastructure/constructs/api.py
+++ b/infrastructure/constructs/api.py
@@ -1,4 +1,5 @@
 from aws_cdk import (
+    RemovalPolicy,
     Tags,
     aws_iam,
     aws_lambda_python_alpha,
@@ -9,6 +10,7 @@ from aws_cdk import (
 )
 from constructs import Construct
 
+from geostore.environment import is_production
 from geostore.resources import Resource
 
 from .common import grant_parameter_read_access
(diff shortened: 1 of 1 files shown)


infrastructure/constructs/api.py
```diff
@@ -67,6 +69,11 @@ def __init__(  # pylint: disable=too-many-locals
             users_role=api_users_role,
             botocore_lambda_layer=botocore_lambda_layer,
         )
+
+        if is_production():
```
l0b0, line 73: Should this and the other place use `if not is_production():`?

Jimlinz:
Relevant discussion (snippets) from Slack:

---
_Victor Engmark_
https://github.com/aws/aws-cdk/issues/11549 was closed by https://github.com/aws/aws-cdk/pull/21113, but someone else mentioned https://github.com/aws/aws-cdk/issues/21804 applying to Lambda function log groups in particular (associated PR). There's a workaround in that thread. (edited) 

---

_Jim Gan_
```
new logs.LogGroup(this, 'log-group', {
  logGroupName: `/aws/lambda/${onEvent.functionName}`,
  removalPolicy: RemovalPolicy.DESTROY,
  retention: logs.RetentionDays.ONE_DAY,
})
```
So this is essentially "guessing" the log group name the lambda function is going to create, and setting the removal policy to destroy (as a separate entity (for a lack of a better term) to the entity that is creating the lambda function). In our case, we just apply the ci seed and the lambda name. If I am reading this correctly.

---

_Jim Gan_
Each ci run creates around 14 log groups , some with fairly easy guessable name (e.g. /aws/lambda/ci3511252566-dataset-versions), others a bit tricky (e.g. /aws/lambda/ci3511252566-geostore-processingGetUploadStatusFun-5J8msKxxAlty). Not sure if there is an easy way to determine how cdk generates 5J8msKxxAlty suffix - I will dig a bit to see if there is an easy way to work this out or tack this on to removal.
This workaround do feel a bit messy though.

---

Victor Engmark
Yeah, we might just end up shelving this as blocked until that PR (or something like it) gets merged if the workaround is too ugly.

Jimlinz:
Workaround implemented in #2338

PR #2336, opened 2022-11-22
feat: Ensure only one GitHub action workflow is running

### LINKED ISSUE #2290: Cancel current GitHub Action workflow (if still running) when subsequent commits are made
### Enabler

So that we reduce GitHub Actions minutes and costs, we want to ensure that any currently running GitHub Action workflow for a particular Pull Request is cancelled when another commit is made at a later date (i.e. within 40min or before the original run is complete).

When a Pull Request is opened, a GitHub Action workflow is triggered as part of CI. This process is about 30 to 40 minutes in total. Occasionally, we discovered that the changes made in this Pull Request isn't complete or that we may have missed something (or another test has failed). We make the appropriate changes and push the code up. This triggers yet another GitHub Action workflow run. The previous workflow is left running until completion. This seems rather wasteful as the previous workflow is no longer relevant (in most cases).

Reference:
https://stackoverflow.com/a/72408109

#### Acceptance Criteria

-  [ ] Ensure only one GitHub Action workflow run per Pull Request per branch is allowed at any one time. 
-  [ ] Previous workflow run cancelled when new commits are made (or when new workflow is triggered).
-  [ ] Ensure it doesn't affect workflow in another branch (i.e. it doesn't end up cancelling workflow on a different pull request)
-  [ ] Ensure proper clean up is established when an existing run is cancelled (i.e. no artifacts left behind in s3 etc).

#### Additional context

What happens if the step function is in mid run when the job is cancelled? Would it run to completion and output the files onto a s3 bucket that may have been deleted prior? Would it fail?

Jimlinz:
Geostore CI takes a while to complete. In cases where a new commit is pushed to an existing pull request branch before the previous GitHub action workflow completes, a new workflow is triggered; leaving the old workflow to run its course even though it is no longer relevant (in most cases). This PR limits workflow concurrency by cancelling previous workflow.

This resolves #2290 

References:
https://docs.github.com/en/actions/using-jobs/using-concurrency
https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre/72408109#72408109

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [master]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   CodeQL-Build:
     runs-on: ubuntu-20.04
--- a/.github/workflows/package-cli.yml
+++ b/.github/workflows/package-cli.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ubuntu-20.04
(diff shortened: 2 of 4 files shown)


l0b0 (approved):
Nice bit of research!

Issue #2319, opened 2022-11-17
Cull CI log groups periodically


Jimlinz:
Our CI has seed prefixes. This is useful in many ways, especially for troubleshooting, as well as setting up the stage for running pytest tests in parallel. However, one downside is that it creates a handful of new log groups for each seed / CI run (one for each lambda per CI run). This adds up very quickly. 

While there is an option to set individual log entry to expire, there isn't an option to delete the log group itself within the geostore stack. AFAIK log group `RemovalPolicy.DESTROY` isn't an option (see https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda_python_alpha/PythonFunction.html). 

This isn't a problem in production, but it is an issue in CI, where CloudWatch logs are pretty much unusable past 10,000 groups. 

We should look at various options to keep the number of log groups in check in CI. This would not only keep things tidy, keep costs down (less logs), but also allows us to actually view and search the logs when troubleshooting. 

Options:
- GitHub actions to cull logs on a schedule
- Lambda function to cull logs on a schedule
- Find a way to clean up after itself (delete own log groups) during teardown (but would make the logging less useful as they are gone before one could look at it for troubleshooting purposes)
- ???
- Profit

Please feel free to add your thoughts and suggestions below

l0b0:
We should ideally have tests for what goes into the logs (and we do for a bunch of things), so the logs in the CI instances shouldn't really provide any new info we couldn't gather from the CI output. So I'd be in favour of tearing these down after the tests.

billgeo:
I agree. And if there was any artifacts we need after CI, should be kept as Github artifacts which can then be configured to be deleted after some set period.

Jimlinz:
Merged: https://github.com/linz/geostore/pull/2338

billgeo:
This has been done. And all old log groups will eventually be removed.

PR #2338, opened 2022-11-22
feat: Delete CloudWatch Log Groups over a week old

Jimlinz:
There is no mechanism in place to clean up CloudWatch Log Groups created by Geostore. While the logs itself can be set to expire, the Log Groups are kept forever. This presents a problem in CI. 

There isn't an elegant way to delete log groups within CDK on stack teardown currently (this may change in the future). 

This PR runs `aws logs delete-log-group` command at the end of the CI to cull log groups older than 1 week old (from previous CI runs). **This is a temporary work around until a better solution is found.** 

Note: A `--no-paginate` option is added to limit the number of queries (~50). This prevents rate limiting as well as stops a potentially endless number of results being returned by AWS (which would result in a very log running workflow). This means only a maximum number of 50 log groups are removed each workflow run.

A workaround / partial fix to #2319

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -176,6 +176,13 @@ jobs:
           aws cloudformation delete-stack --stack-name ci${GITHUB_RUN_ID}-geostore
         if: always()
 
+      - name: Delete CloudWatch Log Groups over a week old
+        run: |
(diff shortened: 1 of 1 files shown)


l0b0 (commented):
Code looks good, but I'd want more information in the commit message. Basically all the info in the PR description, ideally including links to relevant issues like https://github.com/aws/aws-cdk/issues/21804.

l0b0 (approved):

PR #2339, opened 2022-11-23
feat: Enable SNS Topic TLS encryption

Jimlinz:
`cdk-nag` recommends enabling SNS topic TLS encryption, as below: 

```
AwsSolutions-SNS2: The SNS Topic does not have server-side encryption enabled. Server side encryption adds additional protection of sensitive data delivered as messages to subscribers.

AwsSolutions-SNS3: The SNS Topic does not require publishers to use SSL. Without HTTPS (TLS), a network-based attacker can eavesdrop on network traffic or manipulate it, using an attack such as man-in-the-middle. Allow only encrypted connections over HTTPS (TLS) using the aws:SecureTransport condition and the 'sns: Publish' action in the topic policy to force publishers to use SSL. If SSE is already enabled then this control is auto enforced.
```

This PR proposes to resolve this by applying TLS encrpytion (using AWS managed key) to geostore SNS topic.

According to https://aws.amazon.com/kms/pricing/, you are not charged for the following:
Creation and storage of AWS managed or AWS owned KMS keys. These keys are automatically created on your behalf when you first attempt to encrypt a resource in an AWS service that integrates with AWS KMS. You can neither manage the lifecycle or access permissions on AWS managed keys.

IAM policy for creating and managing keys are needed as per AWS documentation 
- https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-default.html
- https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_kms/README.html


l0b0 (commented):
LGTM, but seems to fail tests

PR #2361, opened 2022-11-25
fix: Delete vended log groups in ci and non-prod

Jimlinz:
Vended log groups created by state-machine should be destroyed on stack teardown in ci and non prod environments. This is to keep the number of log groups manageable (especially since new log groups are created with each ci seed).

### DIFF
--- a/infrastructure/constructs/processing.py
+++ b/infrastructure/constructs/processing.py
@@ -1,5 +1,6 @@
 from aws_cdk import (
     Duration,
+    RemovalPolicy,
     Tags,
     aws_dynamodb,
     aws_iam,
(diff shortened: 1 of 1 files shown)


infrastructure/constructs/processing.py
```diff
@@ -82,7 +83,7 @@ def __init__(
         datasets_table: Table,
         git_commit_parameter: aws_ssm.StringParameter,
     ) -> None:
-        # pylint: disable=too-many-locals, too-many-statements
+        # pylint: disable=too-many-locals, too-many-statements, too-complex
```
l0b0, line 86: Looks like this method should be pulled apart. Would you be OK to do that in a separate branch?

l0b0 (approved):

infrastructure/constructs/processing.py
Jimlinz: Yes I was thinking the same!

PR #2395, opened 2022-11-30
feat: Use commit date for nixpkgs.json name

l0b0:
Avoids using a name dependent on when the script is run.

Also swap diff inputs. Shows the more natural diff with the original contents on the left and the new contents on the right.

## Reference


Jimlinz (approved):
It took me several read to follow what is going on, but maybe that is just my lack of experience with bash.

PR #2468, opened 2022-12-15
refactor: Remove botocore lambda layer

Jimlinz:
Python lambda function comes with botocore by default. Adding a lambda layer to include this seems redundant. Ref: https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html


Jimlinz:
This is still needed as a safe guard against AWS changing the botocore version without warning, which would make the things very difficult to troubleshoot. In any case, the botocore version provided by AWS is old. This is a shame as removing this would vastly simplify our code base. 

Interestingly,  removing botocore layer didn't really shave much time off deployment.

PR #2475, opened 2022-12-15
fix: Show annotated git tag

Jimlinz:
An error message matching `fatal: no tag exactly matches <sha1 #>` is thrown when running `cdk synth` or `cdk deploy` locally. This can be misleading as it doesn't explain why a fatal error is thrown, and by what. This problem is caused by `git tag` creating non-annotated tag by default. Rather than calling git commands using Popen, GitPython module provides a much elegant way to interact with git repositories.

Ref: 
https://stackoverflow.com/a/41210204
https://stackoverflow.com/a/31534834
https://stackoverflow.com/a/32524783

### DIFF
--- a/infrastructure/constructs/version.py
+++ b/infrastructure/constructs/version.py
@@ -1,12 +1,11 @@
-from subprocess import PIPE, Popen
+from git import Repo  # type: ignore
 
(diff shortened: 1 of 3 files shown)


Jimlinz:
I have no idea why codeql is unhappy.

PR #2504, opened 2022-12-21
feat: Run `twine check` on package files

l0b0:
## Reference


l0b0:
Might be redundant since it's passing with no changes. Maybe Poetry already takes care of this? Closing to avoid introducing additional dependencies, possibly for no reason.

PR #2516, opened 2022-12-22
refactor: Move Python library packaging out of lambda container

Jimlinz:
The current way of packaging python libraries seems a bit wasteful and inefficient. Each lambda within geostore calls `bundle.bash` during `cdk deploy`; which, in turn, pulls a stock lambda container, install pip and poetry, before installing the required packages for the lambda function (rinse and repeat) . This slows down deployment and is incredibly inefficient. A way to work around this is to create a separate temporary directory (i.e. .lambda_out), install the packages locally with a cache directory (so common packages can be reused), before copying the files into the lambda container.


infrastructure/constructs/bundled_code.py
```diff
@@ -1,14 +1,73 @@
...
 
 
-def bundled_code(directory: str) -> aws_lambda.Code:
+def poetry_export_extras(packaging_directory: str, lambda_directory: str) -> None:
+    # There isn't an elegant way of getting poetry to install package dependencies in a bespoke
+    # target lambda_directory within Python, so we export a requirements file and install using pip.
+    # This has been raised and discussed by the community as below:
+    # https://github.com/python-poetry/poetry/issues/1937
+
+    export_extras = run(
```
Jimlinz, line 18: It would be nice to not have to call a poetry subprocess. While it is possible to do `poetry add poetry` and call `import poetry` from within, I have yet to find any documentation or example around this on the Internet. Like `pip` (documentation below), I suspect this is a bad idea. Any code change within `poetry` could make things difficult to troubleshoot...

infrastructure/constructs/bundled_code.py
```diff
@@ -1,14 +1,73 @@
...
+        ["poetry", "export", f"--extras={lambda_directory}", "--without-hashes"],
```
l0b0, line 19: This *should* work without `--without-hashes` since https://github.com/python-poetry/poetry/issues/1584 was fixed.

infrastructure/constructs/bundled_code.py
```diff
@@ -1,14 +1,73 @@
...
+        check=True,
+        stdout=PIPE,
+    ).stdout
+    # Remove botocore as this is already installed in the lambda layer
+    export_extras = sub(b"botocore==.*\n", b"", export_extras)
+
+    with open(
+        f"{packaging_directory}/{lambda_directory}/requirements.txt", "wb"
+    ) as lambda_requirements:
+        lambda_requirements.write(export_extras)
```
l0b0, line 29: You should be able to `--output=OUTPUT` to save to this file in the `poetry` command.

infrastructure/constructs/bundled_code.py
```diff
@@ -1,14 +1,73 @@
...
+    # This has been raised and discussed by the community as below:
+    # https://github.com/python-poetry/poetry/issues/1937
+
+    export_extras = run(
+        ["poetry", "export", f"--extras={lambda_directory}", "--without-hashes"],
+        check=True,
+        stdout=PIPE,
+    ).stdout
+    # Remove botocore as this is already installed in the lambda layer
+    export_extras = sub(b"botocore==.*\n", b"", export_extras)
```
l0b0, line 24: Is it possible to achieve the same with `--without=WITHOUT` and a dependency group which contains only botocore?

infrastructure/constructs/bundled_code.py
```diff
@@ -1,14 +1,73 @@
...
+    check_call(
+        [
+            executable,
+            "-m",
+            "pip",
+            "install",
+            "--no-deps",
+            "--quiet",
+            f"--cache-dir={packaging_directory}/cache",
+            f"--requirement={packaging_directory}/{lambda_directory}/requirements.txt",
```
l0b0, line 47: Can we pass the requirements as a string argument rather than a file? Then we won't need `requirements.txt` at all.

Also, `requirements.txt` should probably not be in the same directory we're using as an install target, since it'll be packaged with our Lambdas and include a file which is not needed.

infrastructure/constructs/bundled_code.py
```diff
@@ -1,14 +1,73 @@
...
+    ) as lambda_requirements:
+        lambda_requirements.write(export_extras)
+
+
+def pip_install_requirements(packaging_directory: str, lambda_directory: str) -> None:
+    # Documentation recommend against calling pip internal api; rather, via command line
+    # https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program
+
+    # Target already exists warning is discussed here
+    # https://github.com/pypa/pip/issues/8063
```
l0b0, line 37: This should no longer happen if we pass requirements as arguments, or if we create `requirements.txt` in some other directory.

infrastructure/constructs/bundled_code.py
```diff
@@ -1,14 +1,73 @@
...
     bundling_options = BundlingOptions(
         image=PYTHON_RUNTIME.bundling_image,  # pylint:disable=no-member
-        command=[join(BACKEND_DIRECTORY, "bundle.bash"), directory],
+        command=[
+            "bash",
+            "-c",
+            "mkdir --parents /asset-output/geostore/$1 && \
+                  cp --archive --update $0/$1/* /asset-output/ && \
+                  cp --archive --update /asset-input/geostore/*.py /asset-output/geostore/ && \
+                  cp --archive --update /asset-input/geostore/$1 /asset-output/geostore/ ",
```
l0b0, line 68: This line can be merged with the previous, since they both have the same target.

infrastructure/constructs/bundled_code.py
```diff
@@ -1,14 +1,73 @@
...
+    poetry_export_extras(packaging_directory, lambda_directory)
+    pip_install_requirements(packaging_directory, lambda_directory)
     bundling_options = BundlingOptions(
         image=PYTHON_RUNTIME.bundling_image,  # pylint:disable=no-member
-        command=[join(BACKEND_DIRECTORY, "bundle.bash"), directory],
+        command=[
+            "bash",
+            "-c",
+            "mkdir --parents /asset-output/geostore/$1 && \
+                  cp --archive --update $0/$1/* /asset-output/ && \
```
l0b0, line 66: The variables should be quoted here, so the third argument would probably end up as `"${0}/${1}/"*`.

infrastructure/constructs/bundled_code.py
Jimlinz: Updated to parse this as a string rather than `requirements.txt` as per suggestion below. Removing `--without-hashes` complicates how this is parse to `pip` (line break).

infrastructure/constructs/bundled_code.py
Jimlinz: This isn't possible from what I can tell.

infrastructure/constructs/lambda_endpoint.py
```diff
@@ -31,7 +32,7 @@ def __init__(
             handler=f"{BACKEND_DIRECTORY}.{package_name}.entrypoint.lambda_handler",
             runtime=PYTHON_RUNTIME,
             timeout=DEFAULT_LAMBDA_TIMEOUT,
-            code=bundled_code(package_name),
+            code=bundled_code(package_name, PACKAGING_DIR),
```
l0b0, line 35: Passing a constant around is usually unnecessary. Normally you'd just import a constant where it's used and reference it directly.

infrastructure/application_stack.py
```diff
@@ -103,3 +105,5 @@ def __init__(self, scope: constructs.Construct, stack_id: str) -> None:
             OpenTopography(
                 self, "opentopography", env_name=env_name, storage_bucket=storage.storage_bucket
             )
+
+        atexit.register(delete_packaging_dir)
```
l0b0, line 109: Is it possible to do this within `bundled_code`? It would be nice to keep it together with the code which creates it. I'm not sure whether registering it more than once would be a problem, though.

infrastructure/constructs/lambda_config.py
```diff
@@ -11,3 +16,9 @@
     RETENTION_DAYS = aws_logs.RetentionDays.ONE_YEAR
 else:
     RETENTION_DAYS = aws_logs.RetentionDays.THREE_MONTHS
+
+PACKAGING_DIR = tempfile.mkdtemp(dir=BACKEND_DIRECTORY, prefix=".lambda_out_")
+
+
+def delete_packaging_dir(packaging_dir: str = PACKAGING_DIR) -> None:
+    shutil.rmtree(packaging_dir)
```
l0b0, line 24: Since this is a single line we could possibly inline it as `lambda: shutil.rmtree(PACKAGING_DIR)`.

infrastructure/constructs/lambda_config.py
```diff
@@ -11,3 +16,9 @@
     RETENTION_DAYS = aws_logs.RetentionDays.ONE_YEAR
 else:
     RETENTION_DAYS = aws_logs.RetentionDays.THREE_MONTHS
+
+PACKAGING_DIR = tempfile.mkdtemp(dir=BACKEND_DIRECTORY, prefix=".lambda_out_")
+
+
+def delete_packaging_dir(packaging_dir: str = PACKAGING_DIR) -> None:
```
l0b0, line 23: This function is never called with `packaging_dir`, so we might as well remove the argument.

infrastructure/application_stack.py
Jimlinz: This isn't possible as far as I can tell. `bundled_code` returns `aws_lambda.Code.from_asset(path=".", bundling=bundling_options)` where `bundling_options` is called. We would prematurely delete `PACKAGING_DIR` before it gets called / used.  

We also package lambda at a several places. Cleaning up at `bundled_code` would mean we recreate `PACKAGING_DIR` again over and over for each lambda?

Can you think of a more elegant way to approach this?

infrastructure/constructs/lambda_endpoint.py
Jimlinz: Good spotting. This was a hang over from previous attempt. Have fixed.

l0b0: This also applies to `pip_install_requirements`.

infrastructure/application_stack.py
l0b0: > This isn't possible as far as I can tell. `bundled_code` returns `aws_lambda.Code.from_asset(path=".", bundling=bundling_options)` where `bundling_options` is called. We would prematurely delete `PACKAGING_DIR` before it gets called / used.

Based on a quick scan of [the docs](https://docs.python.org/3/library/atexit.html), I don't think that's correct:

>  Functions thus registered are automatically executed upon normal interpreter termination.

So it should only be run once the overarching Python process finishes. Which, as far as I can tell, should result in the same behaviour as the original code (with the exception of `rmtree` running multiple times for the same directory).

Jimlinz: Interesting. This does work, it seems. Although it will throw a bunch of errors like so:

```Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/usr/lib/python3.9/shutil.py", line 722, in rmtree
    orig_st = os.lstat(path)
FileNotFoundError: [Errno 2] No such file or directory: 'geostore/.lambda_out_5y5bg5o9'
```

due to `rmtree` being called several times over and over.

Bit of a tradeoff in terms between code tidiness vs error thrown (which could be hard to trace in the future)?

If we  only call `atexit.register(lambda: shutil.rmtree(PACKAGING_DIR))` once in `application_stack.py`, we could put a short one line comment like `# cleaning up lambda packaging directory in bundled_code`, where it would be obvious, vs trying to track down `FileNotFoundError: [Errno 2] No such file or directory: 'geostore/.lambda_out_5y5bg5o9'` for each deployment. This could be confusing for anyone new coming into the code?

l0b0: We could use [`ignore_errors=True`](https://docs.python.org/3/library/shutil.html#shutil.rmtree), although that would also silence other errors. So yeah, maybe a single execution is the way to go. Up to you.

infrastructure/constructs/lambda_config.py
```diff
@@ -11,3 +15,5 @@
     RETENTION_DAYS = aws_logs.RetentionDays.ONE_YEAR
 else:
     RETENTION_DAYS = aws_logs.RetentionDays.THREE_MONTHS
+
+PACKAGING_DIR = tempfile.mkdtemp(dir=BACKEND_DIRECTORY, prefix=".lambda_out_")
```
l0b0, line 19: This being at the root of the file means the directory will be created at import time, which isn't ideal. Anything importing this file will create this directory.

infrastructure/application_stack.py
Jimlinz: > Actually, could we register the handler where we create the packaging directory?

Good point.

infrastructure/constructs/lambda_config.py
Jimlinz: Have turned this into a function, along with `atexit.register` to remove the directory as per comment above.

l0b0 (approved):

l0b0 (approved):

PR #2609, opened 2023-01-11
refactor: Simplify Dockerfile

Jimlinz:
Existing implementation is slightly bloated. It pulls down an ubuntu image and proceed to run `apt-get update`, as well as install `pip `and `poetry` before copying the required files over. This PR proposes to export pull a lightweight `python:3.9-slim` image and install the required packages using pip before copying the required files over.

Ref: https://stackoverflow.com/a/48562835


geostore/Dockerfile
```diff
@@ -1,26 +1,22 @@
...
-
-COPY geostore/pip.txt geostore/poetry.txt poetry.lock poetry.toml pyproject.toml /opt/
-WORKDIR /opt
-RUN apt-get update \
-    && apt-get install --assume-yes --no-install-recommends \
-    curl python-is-python3 python3-pip \
-    && rm -rf /var/lib/apt/lists/* \
-    && pip install --no-cache-dir --requirement=pip.txt \
-    && pip install --no-cache-dir --requirement=poetry.txt
+FROM python:3.9-slim as build
```
l0b0, line 1: Can we get the version from the `.python-version` file? One less thing to keep in sync.

geostore/Dockerfile
```diff
@@ -1,26 +1,22 @@
...
+
+WORKDIR /opt/
```
l0b0, line 3: It looks like the only command which uses a relative path is the `pip install`. We should probably either change all the paths to be relative to `/opt` or remove this line.

geostore/Dockerfile
```diff
@@ -1,26 +1,22 @@
...
+
 ARG task
-RUN poetry install --extras=${task} --no-root --only=main
+ARG packaging
 
+RUN python -m venv /opt/.venv
+ENV PATH="/opt/.venv/bin:$PATH"
```
l0b0, line 9: I think this is only needed so that the `pip` command below runs the right one. So changing that line to `RUN /opt/.venv/bin/pip […]` would be consistent with `ENTRYPOINT` below, and would remove the need for this line.

infrastructure/constructs/task_job_definition.py
```diff
@@ -24,9 +26,19 @@ def __init__(
         else:
             batch_job_definition_memory_limit = 500
 
+        check_output(
```
l0b0, line 29: This should probably be `check_call` since we're not doing anything with the output.

l0b0 (approved):

